PR #23, opened 2017-10-30
add macOS start script

konding:


robertmuehsig:
For normal projects we try to make "dotnet electronize start" as clever and fast as possible, but I know that @GregorBiswanger uses the start.cmd script at least for development of the api, so this file here is very helpful for development. 

The big goal however would be to ensure that dotnet electronize start, right @GregorBiswanger ?

rhwy:
for any help it could be, I can confirm that `dotnet electronize start` doesn't work on mac because it looks for electron.cmd (in mac the electron app is just `electron`)

robertmuehsig:
I will merge this change, but both "start"-shell scripts will be removed in the future. In theory dotnet electronize start should do exactly the same (plus more) steps done in these files.

But unfortunately electronize start is currently windows only, so I will merge this file, fix electronize start and if everything worked as expected remove the shell scripts again.

Hope this makes sense!

PR #31, opened 2017-11-03
Should never use async void

tebeco:
Should never use async void

I tried other scenario but that seems to endup in deadlock or the WebHostBuilder is not happy
I ended up avoiding 

Not working so far:
```csharp
public async Task Configure(IApplicationBuilder app, IHostingEnvironment env)
{
  await Electron.WindowManager.CreateWindowAsync()
}
```

```csharp
public void Configure(IApplicationBuilder app, IHostingEnvironment env)
{
  Electron.WindowManager.CreateWindowAsync().Start()
}
```

```csharp
public void Configure(IApplicationBuilder app, IHostingEnvironment env)
{
  Electron.WindowManager.CreateWindowAsync().RunSynchronously()
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ public static IWebHost BuildWebHost(string[] args)
 Open the Electron Window in the Startup.cs file: 
 
 ```csharp
-public async void Configure(IApplicationBuilder app, IHostingEnvironment env)
+public void Configure(IApplicationBuilder app, IHostingEnvironment env)
 {
     if (env.IsDevelopment())
     {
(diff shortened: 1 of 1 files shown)


yeskunall:
I'm just going to place this quote from [this](https://msdn.microsoft.com/en-us/magazine/jj991977.aspx) article here in case anyone is wondering why this is useful:

> Async void methods have different error-handling semantics. When an exception is thrown out of an async Task or async Task<T> method, that exception is captured and placed on the Task object. With async void methods, there is no Task object, so any exceptions thrown out of an async void method will be raised directly on the SynchronizationContext that was active when the async void method started.

... also _**because**_ `Electron.WindowManager.CreateWindowAsync` seems to be a CPU-bound task,  spawning the work off on another thread with `Task.Run` will ensure responsiveness.

tebeco:
Well the question is
should/does AspnetCore  support Async scenario while `Configure() `

robertmuehsig:
The normal Configure() is not async as well, so it seems your suggestion is "better" with the Task.Run...

PR #32, opened 2017-11-03
use XmlWritter to avoid serialising Xml Declaration in csproj

tebeco:
the idea is to have the minimum amount of change detected in csproj

`<?xml version ...?>`

as never been part of csproj

### DIFF
--- a/ElectronNET.CLI/Commands/InitCommand.cs
+++ b/ElectronNET.CLI/Commands/InitCommand.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using System.Runtime.InteropServices;
 using System.Text;
 using System.Threading.Tasks;
+using System.Xml;
 using System.Xml.Linq;
 
 namespace ElectronNET.CLI.Commands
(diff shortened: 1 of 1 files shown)


ElectronNET.CLI/Commands/InitCommand.cs
```diff
@@ -147,10 +148,19 @@ private static bool EditCsProj(string projectFile)
...
+                };
+                using (XmlWriter xw = XmlWriter.Create(stream, xws))
+                {
+                    xmlDocument.Save(xw);
+                }
 
-                Console.WriteLine($"{ConfigName} added in csproj!");
             }
+
+            Console.WriteLine($"{ConfigName} added in csproj!");
```
robertmuehsig, line 163: Mh - why did you move the Console.WriteLine out of the if(...) section? In theory you can run "init" any time, but we only will add our file once - thats why it is inside the if statement

tebeco: i wanted to move it out for the top level using block in fact i might have done that too fast (no more computer will update tomorrow)

tebeco: it was not in any if() section right ? just the last line of the using()
now it is not

robertmuehsig: AHHH  - whoops - sorry! You are completely right! Looks good!

robertmuehsig:
Will merge it this weekend and release a new NuGet package!

robertmuehsig:
Thanks! Thats great!

PR #35, opened 2017-11-03
Update ElectronNET.WebApp.csproj

engineers-tools:
Changed reference to 0.0.5. Otherwise it fails when building example WebApp

### DIFF
--- a/ElectronNET.WebApp/ElectronNET.WebApp.csproj
+++ b/ElectronNET.WebApp/ElectronNET.WebApp.csproj
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk.Web">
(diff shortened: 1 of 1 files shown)


robertmuehsig:
The reason why it fails is, that you the local /artifacts-nuget folder is propably empty and NuGet will fall back to NuGet.org, where version 0.0.5 is published. 
The WebHost in this repo is for our development and our development bits are versioned as 1.0.0.

Maybe if you run buildAll.cmd before anything else it should work. 

The idea of the WebHost in this repo is to support our development effort - in this case we can't link the version from NuGet.org :)

robertmuehsig:
Hope this makes sense for you. Our documentation sucks currently - maybe you could try to use the buildAll.cmd and it should work.

engineers-tools:
Thanks for the clarification. My apologies for the unnecessary pull-request.
I'd just say this project is fantastic. I'd been looking for something like this for a while and this is the perfect solution. If there's anything I can do to help, please ask.

PR #38, opened 2017-11-04
Move to generic RIDs

Petermarcu:
In .NET Core 2.0, generic RID's should be used to enable broadest reach for the resulting published app. This change just moves all the scripts/projects to use that instead.

### DIFF
--- a/ElectronNET.CLI/Commands/BuildCommand.cs
+++ b/ElectronNET.CLI/Commands/BuildCommand.cs
@@ -41,7 +41,7 @@ namespace ElectronNET.CLI.Commands
                 switch (desiredPlatform)
                 {
                     case "win":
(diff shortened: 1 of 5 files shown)


robertmuehsig:
Uhh - does that mean with one build we could run (at least the .NET Core part) on any Windows Version that supports .NET Core 2.0? e.g. I build it on a Windows 10 machine, but the target platform is Windows 8.1 - that will work, right?

Petermarcu:
That's right. One .net core app runs on Win7+. I don't know enough about electron to know if the same is true there.

robertmuehsig:
Thats awesome! The docs should point this more out! Thanks!

Petermarcu:
Agreed, I'll probably make some PR's to the docs.

PR #40, opened 2017-11-04
Change ElectronNET.API ref to 0.0.5

Petermarcu:
The current reference is to a package that doesn't exist on NuGet. If this isn't the right fix, maybe the 1.0.0 packages is missing on NuGet.org?


robertmuehsig:
This is "by design" - details here: https://github.com/ElectronNET/Electron.NET/pull/35

PR #54, opened 2017-11-13
App cancellation token

danielmarbach:
This would add cancellation token support to the async App methods

### DIFF
--- a/ElectronNET.API/App.cs
+++ b/ElectronNET.API/App.cs
@@ -464,40 +464,49 @@ namespace ElectronNET.API
         /// The current application directory.
         /// </summary>
         /// <returns></returns>
-        public async Task<string> GetAppPathAsync()
+        public async Task<string> GetAppPathAsync(CancellationToken cancellationToken = default(CancellationToken))
         {
+            cancellationToken.ThrowIfCancellationRequested();
+
             var taskCompletionSource = new TaskCompletionSource<string>();
-
-            BridgeConnector.Socket.On("appGetAppPathCompleted", (path) =>
+            using(cancellationToken.Register(() => taskCompletionSource.TrySetCanceled()))
             {
-                BridgeConnector.Socket.Off("appGetAppPathCompleted");
-                taskCompletionSource.SetResult(path.ToString());
-            });
+                BridgeConnector.Socket.On("appGetAppPathCompleted", (path) =>
+                {
+                    BridgeConnector.Socket.Off("appGetAppPathCompleted");
+                    taskCompletionSource.SetResult(path.ToString());
+                });
 
-            BridgeConnector.Socket.Emit("appGetAppPath");
+                BridgeConnector.Socket.Emit("appGetAppPath");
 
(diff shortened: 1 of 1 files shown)


ElectronNET.API/App.cs
```diff
@@ -464,40 +464,47 @@ public void Show()
         /// The current application directory.
         /// </summary>
         /// <returns></returns>
-        public async Task<string> GetAppPathAsync()
+        public async Task<string> GetAppPathAsync(CancellationToken cancellationToken = default(CancellationToken))
```
danielmarbach, line 467: With CSharp 7.1 support this could be reduced to `= default` but I didn't want to change too many things at the same time

ElectronNET.API/App.cs
```diff
@@ -464,40 +464,47 @@ public void Show()
...
         {
-            var taskCompletionSource = new TaskCompletionSource<string>();
+            cancellationToken.ThrowIfCancellationRequested();
 
-            BridgeConnector.Socket.On("appGetAppPathCompleted", (path) =>
+            var taskCompletionSource = new TaskCompletionSource<string>();
+            using(cancellationToken.Register(() => taskCompletionSource.TrySetCanceled()))
```
danielmarbach, line 472: I deliberately choose here the non state based overload. This leads to closure allocations but since the BridgeConnector has the same problem I figured it is OK

ElectronNET.API/App.cs
```diff
@@ -464,40 +464,47 @@ public void Show()
...
-                taskCompletionSource.SetResult(path.ToString());
-            });
+                BridgeConnector.Socket.On("appGetAppPathCompleted", (path) =>
+                {
+                    BridgeConnector.Socket.Off("appGetAppPathCompleted");
+                    taskCompletionSource.TrySetResult(path.ToString());
+                });
 
-            BridgeConnector.Socket.Emit("appGetAppPath");
+                BridgeConnector.Socket.Emit("appGetAppPath");
```
danielmarbach: Will do white space cleanup when done

ElectronNET.API/App.cs
```diff
@@ -464,40 +464,47 @@ public void Show()
...
 
-            return await taskCompletionSource.Task;
+                return await taskCompletionSource.Task;
```
danielmarbach: Would you also like me to add the missing `ConfigureAwait`?

robertmuehsig:
Looking good - what do you think @GregorBiswanger?

GregorBiswanger:
@danielmarbach @robertmuehsig I think this is important for a clean and solid code... Thank you @danielmarbach for your support... Can you implement it on the other Electron.cs API classes too? :)

danielmarbach:
App cancellation token ready for review

danielmarbach:
Should or could this be merged?

robertmuehsig:
@GregorBiswanger Bump!

GregorBiswanger:
Thank you @danielmarbach... 👍

Issue #29, opened 2017-11-03
Fork > dotnet restore > fail


tebeco:
Hi,

I Just tried yo fork and restore for a PR but here it how it went :

```
C:\Workspace\GitHub\TeBeCo>git clone https://github.com/tebeco/Electron.NET
Cloning into 'Electron.NET'...
remote: Counting objects: 1442, done.
remote: Compressing objects: 100% (315/315), done.
remote: Total 1442 (delta 268), reused 383 (delta 170), pack-reused 924
Receiving objects: 100% (1442/1442), 3.40 MiB | 4.55 MiB/s, done.
Resolving deltas: 100% (836/836), done.

C:\Workspace\GitHub\TeBeCo>cd Electron.NET
C:\Workspace\GitHub\TeBeCo\Electron.NET>code-insiders .
C:\Workspace\GitHub\TeBeCo\Electron.NET>dotnet restore
  Restoring packages for C:\Workspace\GitHub\TeBeCo\Electron.NET\ElectronNET.WebApp\ElectronNET.WebApp.csproj...
  Restoring packages for C:\Workspace\GitHub\TeBeCo\Electron.NET\ElectronNET.CLI\ElectronNET.CLI.csproj...
  Restoring packages for C:\Workspace\GitHub\TeBeCo\Electron.NET\ElectronNET.API\ElectronNET.API.csproj...
  Restoring packages for C:\Workspace\GitHub\TeBeCo\Electron.NET\ElectronNET.WebApp\ElectronNET.WebApp.csproj...
  Generating MSBuild file C:\Workspace\GitHub\TeBeCo\Electron.NET\ElectronNET.CLI\obj\ElectronNET.CLI.csproj.nuget.g.props.
  Generating MSBuild file C:\Workspace\GitHub\TeBeCo\Electron.NET\ElectronNET.CLI\obj\ElectronNET.CLI.csproj.nuget.g.targets.
  Restore completed in 214.71 ms for C:\Workspace\GitHub\TeBeCo\Electron.NET\ElectronNET.CLI\ElectronNET.CLI.csproj.
  Restore completed in 253.32 ms for C:\Workspace\GitHub\TeBeCo\Electron.NET\ElectronNET.WebApp\ElectronNET.WebApp.csproj.
  Generating MSBuild file C:\Workspace\GitHub\TeBeCo\Electron.NET\ElectronNET.API\obj\ElectronNET.API.csproj.nuget.g.props.
[...]

PR #56, opened 2017-11-14
fix: run CLI dll extension using dotnet cli

tebeco:
see

https://github.com/ElectronNET/Electron.NET/issues/29

### DIFF
--- a/buildAll.cmd
+++ b/buildAll.cmd
@@ -16,10 +16,10 @@ dotnet build
 echo "Invoke electronize build in WebApp Demo"
 
 echo "-- win"
-dotnet electronize build win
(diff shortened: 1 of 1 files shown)


robertmuehsig:
For "local development" with this repository this would be ok I guess. The current solution for our development here is quite frustrating because of the NuGet cache.

But just that we all are on the same side: This is just an issue with our development enviroment. In theory you should be able to run dotnet electron..., because this is the way to consume dotnet cli extensions.

tebeco:
Of course this is only for the "BuilAll.cmd" so this repo only for dev
once the CLI is published, it is either used from within a nuget of an executable in the path,

Still this `buiildall.cmd` has no idea about any of it

robertmuehsig:
Ok - I will merge it tonight, maybe this will also resolve our nuget cache pain 👍

PR #57, opened 2017-11-14
BuildAll sh/cmd changes

robertmuehsig:


robertmuehsig:
directly invoke the builded dlls

Issue #72, opened 2017-12-06
ASP.NET process is spawned with confusing working directory


tboby:
When Electron.NET starts the asp.net application process it doesn't set a cwd. This means the cwd of the running process is used. Confusingly this is the location of `electron.cmd`, which is `app\node_modules\.bin\`.

Would it not make more sense for the asp.net application to start with its own directory as the cwd? This means calls such as `File.ReadAllLines("./appsettings.json")` behave as expected.

Currently that would read `app\node_modules\.bin\appsettings.json` not `app\bin\appsettings.json`.

https://github.com/ElectronNET/Electron.NET/blob/200f51194062a953e1924c309e4de65941f2119b/ElectronNET.Host/main.js#L51-L52

I think the change would be something like:
```
var options = { cwd: path.join(__dirname,'bin') }
apiProcess = process(binFilePath, parameters, options);
```

robertmuehsig:
Mh - I'm not sure what the working directory really is, but at least appSettings should work fine, see this https://github.com/ElectronNET/Electron.NET/issues/42 with this sample commit https://github.com/ElectronNET/Electron.NET/commit/24ef9874716cb5546e65fabf12c5ffb05533bd3f#diff-627e2365a214cf35b5d1d98ef9b1a9c5

robertmuehsig:
But maybe it would be good anyway to set the working directory properly - good catch!

jcphlux:
I am using env.ContentRootPath and env.WebRootPath that might get you the path you need.

robertmuehsig:
It's quite late, sorry, but when did this occure? After you use "dotnet electronize build" or after "dotnet electronize start"?

tboby:
I thought after both, the javascript that starts the backend is the same I think.
I fixed it by manually setting the CWD in the app, but that's not ideal!

PR #101, opened 2018-02-19
https://github.com/ElectronNET/Electron.NET/issues/72

robertmuehsig:
@GregorBiswanger Works on my machine (the XmlDoc cleanup is not important... but resulted in build warning...)

Important part: Setting the current working dir. Any problems with that? Seems "ok"


robertmuehsig:
@GregorBiswanger I will merge this, ok?

PR #107, opened 2018-03-16
WebContent.GetUrl() and LoadURLOptions.ExtraHeaders

ru-sh:
**WebContent.GetUrl()** 
is used to get the Url of the loaded page. It's usefull if a web-server redirects you and you need to know where it redirects. For instance, It's useful in case of Implicit Authorization.

**LoadURLOptions.ExtraHeaders**
is used to setup additional page loading options. For example ExtraHeaders = "pragma: no-cache\n" forces Chrome to load web page without using of the cache.


robertmuehsig:
@GregorBiswanger Merge?

ru-sh:
Yes. Thank you.

Issue #118, opened 2018-04-18
No window on raspberry appearing


80er:
Hi,

followed the example and works on my Win10 machine. Either by command line or when double clicking ElectronNET.Host.exe.

When i deploy it to my Raspi and try to run the ElectronNET.Host simply nothing happens. only around 20 processes are started and no window appears. No exception or error is written on console.
Electron quickstart sample works. Also made sure that in ElectronNet project electron in v 1.7.11 is installed (1.8.x doesn't work at all on Raspi). 
Any hints where i could further check?

80er:
OK. found it. After some digging it appeared that the asp process did not start. Reason is wrong binary format. 
C:\Users\Chef\Source\Repos\RaspiMediaController\RaspiMediaControllerFrontend>dotnet electronize build /target linux /electron-arch armv7l
Build Electron Application...
Build ASP.NET Core App for linux-x64...

Of course linux-x64 is wrong. But how can i force it use the correct one? Using the same target as for dotnet publish (so linux-arm) suddenly builds a 64bit windows executable... Any hints what is the correct command line parameter?

GregorBiswanger:
Hi @80er 

do you use Windows 10 IoT? Or?

80er:
Hi, 
nope. Do not ask why but i want to get this running on a Raspi with latest stretch. ;) 
So it is running a 32bit linux.
Already wrote small services using ASP.NET Core and then used
``
dotnet publish -r linux-arm
``
to generate the binaries. That works perfect. 
I also have a workaround now. After 
``
dotnet electronize build .... 
``
I call again
``
dotnet publish -r linux-arm 
``
and copy the generated binaries into the packages directory created before under resources/app/bin. Then i deploy the binaries to the Raspi and call ElectronNET.Host and it works. But of course would be nicer to have something like
``
dotnet electronize build /target linux-arm ...
``
and that's it.

GregorBiswanger:
@robertmuehsig do you can implement this for Version 0.0.10?

robertmuehsig:
@GregorBiswanger PR merged - there are a couple of other PRs open, maybe take a look and I will try to build something this weekend.

niklr:
After copying the content of "bin\desktop\electron.net.host-linux-x64" to my Raspberry Pi 3 and setting the execute permission with "chmod 755 ./electron.net.host" the following error is returned after executing with "./electron.net.host":

[...]

PR #119, opened 2018-04-19
added support for linux-arm platform

80er:
Hi,

as discussed in the issue #118 . Direct support for linux-arm added. Platform for packager is still linux. electron-arch gets already piped through correctly.


robertmuehsig:
Thanks!

PR #139, opened 2018-06-06
Electron 2.0.2

deinok:


robertmuehsig:
@GregorBiswanger - what do you think?

PR #169, opened 2018-10-02
Add the Host error logging

yamaoto:
I added error logging of executable project. In my case, my project can throws an errors while executing

### DIFF
--- a/ElectronNET.CLI/ProcessHelper.cs
+++ b/ElectronNET.CLI/ProcessHelper.cs
@@ -42,15 +42,15 @@ namespace ElectronNET.CLI
                         // we delegate it to cmd.exe, which runs fine
                         // but we can catch any error here and return
                         // 1 if something fails
-                        if (e != null && string.IsNullOrWhiteSpace(e.Data) == false)
-                        {
-                            if (e.Data.ToLowerInvariant().Contains("error"))
-                            {
-                                returnCode = 1;
-                            }
+                        //if (e != null && string.IsNullOrWhiteSpace(e.Data) == false)
+                        //{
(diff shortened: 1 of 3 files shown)


ElectronNET.CLI/ProcessHelper.cs
```diff
@@ -42,15 +42,15 @@ public static int CmdExecute(string command, string workingDirectoryPath, bool o
                         // we delegate it to cmd.exe, which runs fine
                         // but we can catch any error here and return
                         // 1 if something fails
-                        if (e != null && string.IsNullOrWhiteSpace(e.Data) == false)
-                        {
-                            if (e.Data.ToLowerInvariant().Contains("error"))
-                            {
-                                returnCode = 1;
-                            }
+                        //if (e != null && string.IsNullOrWhiteSpace(e.Data) == false)
```
robertmuehsig, line 45: The idea was to return an error code if something fails. This would break it. Maybe you could tell me your idea behind this change.

ElectronNET.CLI/ProcessHelper.cs
```diff
@@ -61,15 +61,15 @@ public static int CmdExecute(string command, string workingDirectoryPath, bool o
...
```
robertmuehsig, line 64: The idea was to return an error code if something fails. This would break it. Maybe you could tell me your idea behind this change.

GregorBiswanger:
@yamaoto We have with Electron.NET solutions for exception handling:

```
var browserWindow = await Electron.WindowManager.CreateWindowAsync(viewPath);
browserWindow.WebContents.OnCrashed += async (killed) =>{};
browserWindow.OnUnresponsive += async () => {}
```
Please take a look into the Electron.NET Demo App.

PR #178, opened 2018-10-25
Fixing window opening example

pasisavolainen:
Needed this before any window would show up (electron.exe running in background but no window visible)

Env:
Windows 8.1 (9600)
Electron version seems to be 1.7.11.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
     });
 
     // Open the Electron-Window here
-    Task.Run(async () => await Electron.WindowManager.CreateWindowAsync());
+    Task.Run(async () => (await Electron.WindowManager.CreateWindowAsync()).Show());
 }
 ```
 


robertmuehsig:
@GregorBiswanger - is this OK?

GregorBiswanger:
@pasisavolainen You don´t need the Show()-Method.. Only you have an invisible window..
CreateWindowAsync() call inner a default show implementation.. I don´t know, why you need a second call?!

robertmuehsig:
Maybe we should add these lines to the readme sample

https://github.com/ElectronNET/Electron.NET/blob/17e4646d113712bf063596da8bdcd7354c474e0f/ElectronNET.WebApp/Startup.cs#L55-L66

GregorBiswanger:
@robertmuehsig mhhh.. this is not necessary for the default open window feature... the OnReadyToShow is a best practices, if you don´t like a white window for seconds...

@pasisavolainen Works roberts solution better for you?

pasisavolainen:
Does the sample work for you without `.Show()`? I didn't get any visible windows without it and it was confusing when getting started. If Roberts solution achieves visibility-by-default and is the best practice, I'm all for it.

GregorBiswanger:
@pasisavolainen Did you try Robert's code?

PR #195, opened 2018-12-28
Fixes for Issue#155 & AssemblyName != ProjectName

JimmyTheJ:


robertmuehsig:
Is the portscanning too slow or why did you use two arguments: One to set the starting scanning port and another one if this is "fixed". Wouldn't it be "enough" to say: If you specify a port this port should be used?

Otherwise LGTM - Thanks! 

@GregorBiswanger?

JimmyTheJ:
> Is the portscanning too slow or why did you use two arguments: One to set the starting scanning port and another one if this is "fixed". Wouldn't it be "enough" to say: If you specify a port this port should be used?
> 
> Otherwise LGTM - Thanks!
> 
> @GregorBiswanger?

So I want both values in there. The port is the starting port to begin scanning from. If you don't set the fixed property then it will just scan +1-10 from that port. So maybe you want a range other than 8001-8010 this would allow you to do that. The fixed parameter is clearly there to look in that exact port as the electron host port and then that port + 1 is the port traffic is on (Mirroring the 8000, and 8001 as it would have ended up before)

robertmuehsig:
As far as I know (don't laught at me - this part did @GregorBiswanger ) we open 2 ports: One port from the Electron side and the other one for ASP.NET Core.

If you would specify a fixed Port 7777 this would mean Port 7777 is used from the Electron part and Port 7778 is used by ASP.NET Core?

JimmyTheJ:
> As far as I know (don't laught at me - this part did @GregorBiswanger ) we open 2 ports: One port from the Electron side and the other one for ASP.NET Core.
> 
> If you would specify a fixed Port 7777 this would mean Port 7777 is used from the Electron part and Port 7778 is used by ASP.NET Core?

Oh I wouldn't laugh at you, and yes two ports are opened though I'm not 100% sure myself as to what they refer to, but it would stand to reason that you're right.

The port that the world would see (or in this case localhost) is 7778 in your example and if you were using an SPA you would have your configs setup to resolve to that port.

robertmuehsig:
Because we need 2 open ports (one for the electron part and the other one for the ASP.NET Core part) I would prefer the following solution: If you want fixed/predefined ports, both ports would be needed to be specified. 
Another question: 
Where would you specify those arguments? When we package the electron app, there is no easy way to submit arguments, right?

JimmyTheJ:
So in the javascript we run the dotnet core executable with 2 parameters, one for the electron port one for the endpoint port.

I can set it up so when you initialize electron you do "electronize init /electron-port 8000 /access-port 8001" which will setup your two ports in the manifest file and it will just read from that in the javascript. It's half-way to what I've already done. I'll add that in a new commit.

I'll still need the flag for whether we are in "static" mode or "portfinder" mode though I think.

GregorBiswanger:
@JimmyTheJ @robertmuehsig  I think a better solution is, to set the default ports with the appsettings.json.

JimmyTheJ:
@GregorBiswanger how would we have it respect the values in there though ? In main.js the process is called with 2 arguments corresponding to the electron port and the endpoint port that it determines from the default 8000 port + whatever the port scanner finds as next available port.

Do you think we should change the program to not take any arguments from the javascript and just use the appsettings to get those 2 values ?

I'm okay with that for my own implementations, but wouldn't there we people who want the options of doing it either way ?

alexandrereyes:
Hey Guys! I'm trying to set the AssemblyName which is != from ProjectName, so when I run `electronize start` a error message is show:
image

@GregorBiswanger mentioned that setting default port in the appsettings solve the problem, but I tried every method from [this post](https://stackoverflow.com/questions/37365277/how-to-specify-the-port-an-asp-net-core-application-is-hosted-on) and I still couldn't fix this error 😞 

Any clues?

alexandrereyes:
The trick is: `AssemblyName` (on csproj) must be equal `executable` property at `electron.manifest.json`

Issue #181, opened 2018-11-09
Unexpected firewall warnings


torman119:
When we start the final electron app we get 2 firewall warnings popping up (running on windows 10). 
For instance when building and running the Electron.Net sample music player, we get a firewall warning for "ElectronNET.Host" and one for "myelectronmusicplayer.exe". 
Does anybody know why we get these warnings? Why do we get 2 of them? How can we avoid them for our users?

ctrl-d:
Had same issue & stopped looking into it becouse of that

RogerDodger123:
I have solved the firewall prompt when it comes to the api. This was done by just resetting the bind:
```
WebHost.CreateDefaultBuilder(args)
                .UseElectron(args)
                .UseUrls($"http://localhost:{BridgeSettings.WebPort}")
                .UseStartup<Startup>();
```

but the issue lies here:
https://github.com/ElectronNET/Electron.NET/blob/5a676cb76aee7069cfe15fd09aca2c52e5498447/ElectronNET.API/WebHostBuilderExtensions.cs#L34

The api is being bound to `0.0.0.0`

robertmuehsig:
Mh - would " .UseUrls("http://localhost:" + BridgeSettings.WebPort);" trigger the same Firewall Warning? @RogerDodger123 ?

robertmuehsig:
This should be fixed in the next update.

PR #198, opened 2019-01-02
Fix firewall triggers

robertmuehsig:
Trying to fix the Firewall warnings:

https://github.com/ElectronNET/Electron.NET/issues/181

Current status:

Instead of 2 warning, we are down to one. It seems our Socket.IO implementation triggers it via this line:

   io = require('socket.io')(port);

Be aware: This is work in progress - the code now will throw an exception

Maybe interesting links:

https://stackoverflow.com/questions/9419312/socket-io-binding-to-a-specific-ip-address 
https://github.com/adobe/brackets/issues/10182

### DIFF
--- a/ElectronNET.API/WebHostBuilderExtensions.cs
+++ b/ElectronNET.API/WebHostBuilderExtensions.cs
@@ -31,7 +31,7 @@ namespace ElectronNET.API
             if(HybridSupport.IsElectronActive)
             {
                 builder.UseContentRoot(AppDomain.CurrentDomain.BaseDirectory)
-                    .UseUrls("http://0.0.0.0:" + BridgeSettings.WebPort);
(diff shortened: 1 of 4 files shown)


robertmuehsig:
After some fighting against socket.io, I think this should work now.

Instead of those firewall dialogs nothing should appear, because we strictly bind to localhost and only do a port lookup on localhost

These are gone:

image

Everytime this window appears a entry to the firewall settings will be added automatically (doesn't matter if you select anything).

Check here:
image

And search for "Electron" or "ElectronNET" - at least on my machine I had a couple of them. 

After this PR the list should stay clean.

robertmuehsig:
Talked with @GregorBiswanger - he said "OK" ;)

PR #221, opened 2019-02-24
initial update for .net core 3 preview2

ErikApption:
Runs electron.net with .net core 3 preview

### DIFF
--- a/ElectronNET.API/ElectronNET.API.csproj
+++ b/ElectronNET.API/ElectronNET.API.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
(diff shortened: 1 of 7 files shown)


MaherJendoubi:
@ErikApption I wonder to know why not using the latest preview 5 of .NET Core 3

MaherJendoubi:
@robertmuehsig https://dotnet.microsoft.com/platform/dotnet-standard#versions

robertmuehsig:
Thats not what I asked. What do we need to do (during NuGet packaging etc.) to support netcore2.2 and netcore3.0? I know there are some long blogposts out there and when netcore3.0 is "mature" enough we will do this step, but until then I would need some good advice.

MaherJendoubi:
You need to ship 2 different NuGet packages for instance 5.22.13 for netcoreapp2.2 and 5.30.13 for netcoreapp3.0. You could keep the master branch for netcoreapp3.0 and a second branch for netcoreapp2.2.

robertmuehsig:
Checkout https://github.com/ElectronNET/Electron.NET/pull/313

PR #228, opened 2019-03-28
Adds new command line parameters to 'electronize build'

### LINKED ISSUE #227: Package.json location
As a developer, I want to be able to modify the package.json in order to set up the metadata for the electron app.

Therefore the question, why did you decide to compile the package.json and other files into the CLI tool? Every time I want to increase the version number or change anything in the package.json, I need to build the Electron.Net project in order to get the adjusted CLI.
Wouldn't it make more sense to add the package.json to the referencing project and use the CLI version only if the file isn't in place?

gfs:
Fixes #227.

1. Adds four new command line parameters to `electronize build` allowing users to specify the output path, and to provide their own package.json and to specify that they want an `npm install`
- `/package-json` specifies a package.json file the user wishes to use, and sets `/install-modules`
- `/install-modules` specifies that `electronize build` should always `npm install`
- `/relative-path` specifies a relative path to output binaries
- `/absolute-path` specifies an absolute path to output binaries.
2. Fixes a directory change side effect of the `buildReleaseNuGetPackages.cmd` script.
3. Updates the license information in `Package.json` to match the license included in the code (ISC -> MIT)

The path specifications change the path provided to `electron-packager`.

The `package-json` argument overwrites the bundled package.json after laying out the pre-packed files.

I've tested all four commands manually.

I have noticed you have to be very careful to match your custom package.json with the auto deployed package.json. So documentation for these commands should note that. My procedure was to layout the files normally and then edit the package.json setting it aside to provide as an argument.

Update:
After additional work added one more command line parameter.


robertmuehsig:
@GregorBiswanger Didn't you implement something with the custom package.json? Is this merged to master? 
@gfs Gregor is the "mastermind" bind the integration between the node.js world and the .NET world. Is his currently on vacation, but IMHO this looks good. Thanks for your contribution!

gfs:
Thanks @robertmuehsig! Happy to answer any questions here about my changes.

izakntun:
This works with `electronize start` command?

gfs:
I don't believe that this PR added this functionality to the start command. Just build.

PR #249, opened 2019-05-17
Hecatron pull 1

grbd:
Hi,
I've recently made an attempt at redoing the configuration for Electron.Net
I'd imagine this might not make it through as it's a large number of changes.
But i'd be interested in any feedback and I have tried testing all the different options / commands.

## Configuration

I've set things up to use Microsoft.Extensions.Configuration
The way this works is that it first parses multiple sources into key value pairs
then binds those key value pairs to properties of a class.

  * First it attempts to load settings from a json file - "ElectronNetSettings.json" in the current directory
  * These can all be overriden by command line options such as "--npmcommand=npm" or "/npmcommand=npm"
  * The command line is split into arguments and switches, the switches can start with --, -, /
    so --npmcommand -npmcommand /npmcommand are all the same
  * It supports multiple arguments
  * Each command now has a seperate configuration class to store and parse it's settings.

  * I've tried to keep the command line options the same as before but with just additional options
    The one exception might be manually specifying the runtimeid / electronpacker settings in build
    The way that works is that it can auto detect the values, and then the user can just override the settings it pulls back with options if needed

## Json file

For the configuration to give an example looking at the "start" command
it has a setting of builder["start:runtimeid"]
this can be set via "--runtimeid=<value>" on the command line

for the json file it can be picked up via
```
{
  "start": {
    "runtimeid": "<value here>"
  }
}
```

## Other changes

  * Some of the code in the commands has been split into seperate functions just to make it easier to read / follow
  * Help text is now nicely formatted
  * I've not tested it yet but I added a target for the 32bit rasberry pi
  * The build output I've set to be a subdirectory under bin (such as bin/Host for start, bin/desktop for build)
    Usually with C# projects everything built ends up in bin and it's likley the first place someone will look for a built project

I had to remove --production from the npm install during the build command.
A different solution would have been to move some of the devDependencies into dependencies within package.json
I suspect it's working for some users but not others with --production if they have some of the tools installed globaly at the right versions such as typescript

Anyway let me know what you think
Doing it this way might make it easier to add in other commands / options / arguments later on


grbd:
One difference from the original is that you now need to put in an equals sign for the switches
so instead of /target win
it's /target=win or --target=win or -target=win

GregorBiswanger:
I implement currenty electron-builder instead of electron-packager.. we need the electron.manifest.json file for a additional configuration of electron-builder.. I don´t know, how compatibility is it with your work..

grbd:
Looking at the original source code it appears to be using electron-packager for build
https://github.com/ElectronNET/Electron.NET/blob/master/ElectronNET.CLI/Commands/BuildCommand.cs

I've not changed the command being called (I just split the code into separate functions and added more options)

grbd:
I've now rolled in the most recently changes into this pull request for electron-builder
I've made some notes below

## Additional changes

### Build directory

Temp build directory is now under obj
Final build destination is now determined by the electron.manifest.json file - typically bin/Desktop
this matches the current master

### Package.json

I've noticed within Electron.Host/package.json
we need to have "@types/socket.io" added to the production dependencies
it's included in this pull request, but it's not in the current master

  "dependencies": {
    "image-size": "^0.7.4",
    "portscanner": "^2.2.0",
    "socket.io": "^2.2.0",
    "@types/socket.io": "^2.1.2"
  },


### Global install of tsc / typescript

From the looks of things we're using npm --production
and making the assumption the end user has typescript installed globally
I've added the following command into the build code as part of the pull request

```
npm install typescript --global
```

One issue that users have been running into online
is that the directory C:\Program Files (x86)\Microsoft SDKs\TypeScript
is included in most users path by default
which overrides any installed version of tsc with an older one from Microsoft

https://stackoverflow.com/questions/38191287/error-ts5023-unknown-compiler-option-p

I'm in the process of writing some docs so I'll include it as a note in that

GregorBiswanger:
Good to merge @robertmuehsig? It´s your CLI-Part.

robertmuehsig:
@GregorBiswanger Is this still relevant? Many conflicts. 

@grbd My fault - I didn't find enough time to review it and now we have some merge conflicts. If you are still interested to see these changes in a future release, could you merge them with the current master?

grbd:
to be honest I've moved onto https://github.com/chromelyapps/Chromely now since it seems to work quite well and doesn't use as much memory due to it only using libcef

PR #252, opened 2019-05-19
Update README.md

Tum4ik:
Fixed documentation bug.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,8 +41,8 @@ You start Electron.NET up with an `UseElectron` WebHostBuilder-Extension.
 public static IWebHost BuildWebHost(string[] args)
 {
     return WebHost.CreateDefaultBuilder(args)
-        .UseElectron(args)
         .UseStartup<Startup>()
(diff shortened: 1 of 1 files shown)


robertmuehsig:
Both is possible, but I would say this is more convient. Agree @GregorBiswanger ?

GregorBiswanger:
@robertmuehsig both works

Tum4ik:
Unfortunately, for me the current variant is not working. I was not able to start the application.

Tum4ik:
I suggest the most convenient way
```csharp
public static IWebHostBuilder CreateWebHostBuilder(string[] args) =>
      WebHost.CreateDefaultBuilder<Startup>(args)
        .UseElectron(args)
```

robertmuehsig:
I'm ok with this, :shipit: @GregorBiswanger ?

robertmuehsig:
I think this is ok - so I merged it. Sorry for the huge delay & thanks!

PR #266, opened 2019-05-31
Fix ASP.NET backend listening on 0.0.0.0

gfs:
_Why this change?_
Default behavior for windows appears to be to resolve ```localhost``` to ```0.0.0.0``` which causes the Windows firewall to get triggered, but also exposes the ASP.NET backend to remote access which opens up the app to RCE risks unnecessarily.

_Also_
1. Add a comment for why NodeIntegration is true.

2. The version used in the build script is now a variable, only needs to be changed once.

I've validated that with this changed behavior, our App, [Attack Surface Analyzer](github.com/Microsoft/AttackSurfaceAnalyzer), encounters no issues. I'm unable to test it on the WebApp as the version on master currently does not work for me (even without my changes).

### DIFF
--- a/ElectronNET.API/Entities/WebPreferences.cs
+++ b/ElectronNET.API/Entities/WebPreferences.cs
@@ -15,10 +15,10 @@ namespace ElectronNET.API.Entities
         public bool DevTools { get; set; } = true;
 
         /// <summary>
-        /// Whether node integration is enabled. Default is true.
+        /// Whether node integration is enabled. Default is false.
         /// </summary>
-        [DefaultValue(true)]
-        public bool NodeIntegration { get; set; } = true;
+        [DefaultValue(false)]
+        public bool NodeIntegration { get; set; } = false;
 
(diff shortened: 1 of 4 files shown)


GregorBiswanger:
Hi @gfs, thanks for your contribution!

I had deliberately activated NodeIntegration by default.
The problem is that otherwise no IPC communication can take place via the views to controllers. It's a common standard at Electron.NET to want that. I think that would limit the entry and compatibility with existing applications.

What do you think about this?

robertmuehsig:
Preventing the Windows firewall dialog is always great - in a recent PR I thought I fixed the problem, but maybe I was just looking on the nodejs side 🤔
Besides that: I like the env in the build script 👌

gfs:
@GregorBiswanger Understand now.  I think it makes sense to keep your existing default in that case, it would be rough to break a bunch of people. I think it would be good to have a note in the documentation, I'm happy to provide that.

PR #269, opened 2019-05-31
update projects to use netcoreapp2.2

robertmuehsig:
I discovered that our libs are targeting netcoreapp2.1 instead of netcoreapp2.2. 

It should be "ok", but not sure if someone with ASP.NET Core 2.1 is using our lib.


robertmuehsig:
I guess this is a good change and reflects our supported platform statement ;)

PR #286, opened 2019-07-05
Fixed bug: Electron tray icon TypeError

Tum4ik:
The main idea is: the developer of the end-user application should take care of the tray icon path, not the ElectronNET framework.


Tum4ik:
https://github.com/electron/electron/issues/7657

Issue #288, opened 2019-07-11
Build fails with latest electron-builder version


GeorgDangl:
* **Version**:  5.22.13

* **Target**:  Windows

Steps to Reproduce:

Performing a regular build with 

    electronize build /target win

fails since the [BuildCommand](https://github.com/ElectronNET/Electron.NET/blob/master/ElectronNET.CLI/Commands/BuildCommand.cs#L176) seems to pass invalid CLI arguments to `electron-builder`.

I'm getting the following output:

```bash
C:\Project\obj\desktop\win>electron-builder . --config=./bin/electron-builder.json --platform=win32 --arch=x64
cli.js build
Build
Building:
  --mac, -m, -o, --macos   Build for macOS, accepts target list (see
                           https://goo.gl/5uHuzj).                       [array]
  --linux, -l              Build for Linux, accepts target list (see
                           https://goo.gl/4vwQad)                        [array]
  --win, -w, --windows     Build for Windows, accepts target list (see
                           https://goo.gl/jYsTEJ)                        [array]
  --x64                    Build for x64                               [boolean]
  --ia32                   Build for ia32                              [boolean]
  --armv7l                 Build for armv7l                            [boolean]
  --arm64                  Build for arm64                             [boolean]
  --dir                    Build unpacked dir. Useful to test.         [boolean]
  --prepackaged, --pd      The path to prepackaged app (to pack in a
                           distributable format)
C:\Project\obj\desktop\win>
  --projectDir, --project  The path to project directory. Defaults to current
                           working directory.
  --config, -c             The path to an electron-builder config. Defaults to
                           `electron-builder.yml` (or `json`, or `json5`), see
                           https://goo.gl/YFRJOM
Publishing:
  --publish, -p  Publish artifacts, see https://goo.gl/tSFycD
                [choices: "onTag", "onTagOrDraft", "always", "never", undefined]
Other:
  --help     Show help                                                 [boolean]
  --version  Show version number                                       [boolean]
Examples:
  electron-builder -mwl                     build for macOS, Windows and Linux
  electron-builder --linux deb tar.xz       build deb and tar.xz for Linux
  electron-builder --win --ia32             build for Windows ia32
  electron-builder                          set package.json property `foo` to
  -c.extraMetadata.foo=bar                  `bar`
  electron-builder                          configure unicode options for NSIS
  --config.nsis.unicode=false
Unknown arguments: platform, arch
```

[...]

PR #289, opened 2019-07-11
Fix for #288, use a fixed, local version for electron-builder

GeorgDangl:


GregorBiswanger:
I implemented the support for the newest electron-builder.
The next update 5.22.14 will support it.

Issue #301, opened 2019-08-09
BrowserWindow.SetMenu(null) not working


shazha:
Thanks for this great project, it's really helpful  😃

* **Version**:  5.22.14
.Net Core SDK version: latest 2.2 release

* **Target**: Windows 10

The menu bar is supposed to be removed completely by invoking `BrowerWindow.SetMenu(null)` or `Menu.setApplicationMenu(null)`. However, the default menu, which contains standard items such as `File` and `Edit`, is not suppressed and the menu bar is not removed from created BrowserWindow on Windows.

hack2root:
The menu bar is supposed to be removed completely by invoking browerWindow.RemoveMenu();

GregorBiswanger:
Big thanks to @hack2root for your contribution!

This new feature comes with the next version **5.22.15**.

PR #303, opened 2019-08-21
fix issue #301

hack2root:
Fixes issue #301 by adding Electron API [win.removeMenu()](https://electronjs.org/docs/api/browser-window#winremovemenu-linux-windows), extending Electron.NET API by adding `BrowserWindow.RemoveMenu()` method


GregorBiswanger:
Big thanks @hack2root!

PR #311, opened 2019-09-26
Support of AddExtension, RemoveExtension, GetExtensions

Daddoon:
This PR add the support of:

- **BrowserWindow.AddExtension**
- **BrowserWindow.RemoveExtension**
- **BrowserWindow.GetExtensions**

I have tested the code on my side by loading the updated ElectronNET.CLI, and tested with the following sample test in **Startup.cs**:

```csharp
        public async void ElectronBootstrap()
        {
            var browserWindow = await Electron.WindowManager.CreateWindowAsync(new BrowserWindowOptions
            {
                Width = 1152,
                Height = 940,
                Show = false
            });

            await browserWindow.WebContents.Session.ClearCacheAsync();

            browserWindow.OnReadyToShow += async () =>
            {
                browserWindow.Show();

                await Electron.Dialog.ShowMessageBoxAsync(browserWindow, $"Chrome Extensions test");

                //Test for webextensions
                string extensionName = await BrowserWindow.AddExtensionAsync(@"C:/2Bee/Sources/ElectronNET/Electron.NET/ElectronNET.WebApp/wwwroot/chromextensions/iframe_listener");
                await Electron.Dialog.ShowMessageBoxAsync(browserWindow, $"AddExtension: Loaded extension is {extensionName}");

                var extensionsList = await BrowserWindow.GetExtensionsAsync();
                var extensionListString = string.Join(", ", extensionsList.Select(p => $"Name: {p.Name}, Version: {p.Version}"));

                await Electron.Dialog.ShowMessageBoxAsync(browserWindow, $"GetExtensionsAsync: Results are: {extensionListString}");

                BrowserWindow.RemoveExtension(extensionName);

                var extensionsListAfterRemove = await BrowserWindow.GetExtensionsAsync();
                var extensionListAfterRemoveString = string.Join(", ", extensionsListAfterRemove.Select(p => $"Name: {p.Name}, Version: {p.Version}"));

                await Electron.Dialog.ShowMessageBoxAsync(browserWindow, $"GetExtensionsAsync after remove extension: Results are: {extensionListAfterRemoveString}");
            };
            browserWindow.SetTitle(Configuration["DemoTitleInSettings"]);
        }
```

Attached here, a little Chrome extension created from a Mozilla WebExtension i created for my plugin BlazorMobile. Actually, the running code in this extension should not work as it's specific for a special expected environment, but the goal here is just to test that the Chrome extension load in Electron.

[chromextensions.zip](https://github.com/ElectronNET/Electron.NET/files/3658672/chromextensions.zip)

Of course update the **AddExtensionAsync** with a valid extension folder on your side.
The shown dialog should validate the expected behaviors, according to the method call orders and expected return values.


Daddoon:
Just asking a question, do you have any clues about the release pace from your project ?

I mean, i require theses functionnalities for some plugin implementation on my project [BlazorMobile](https://github.com/Daddoon/BlazorMobile) for the Desktop side (using ElectronNET), as this way i would be able to mimic with nearly the exact same code base the code i have done in GeckoView for Android using WebExtensions instead, for managing cleanly some DOM events, like Navigating event from main window, and nested Navigating events from iFrame and subiframes from a same window.

I'm actually trying to implement a kind of minimalist Xamarin.Forms driver for ElectronNET, though this is not the goal, it's just for managing the same code call base from a Xamarin.Forms project, as used on the native side of my plugin.

The binding of Chrome extensions here is mainly to enforce on my side to easily catch / block DOM navigating events, and reforwarding them on a WebView **Navigating** event like a Xamarin.Forms project, to enforce continuation or to cancel depending the developer choice.

I would prefer to continue to use the ElectronNET main repo instead of forking on my own dependencies on my project for having this functionnality.

I may workaround temporiraly by calling remotes IPC from the web page, but it would be a lot more difficult, as it would be the Web application responsible of managing the required information / path etc, instead of the "native" side in C# of the app.

Thanks in advance for any feedback :) !

PR #324, opened 2019-10-15
Added Target option for Start command

### LINKED ISSUE #271: Support electronize start target 32bit
I'm use this library for develop a software but my software depend on some library only run 32bit process. So i use win 10 x64 for develop and when i run this command "electronize start" it will trigger "dotnet publish x64". so i my project only target x86.

callumhutchy:
Uses the same structure as the /target for BuildCommand, changes the 'path' option to follow the same system /path <path>. 
Allows platform targeting when testing applications, which is necessary when using certain libraries.
Fixes #271

### DIFF
--- a/ElectronNET.CLI/Commands/StartElectronCommand.cs
+++ b/ElectronNET.CLI/Commands/StartElectronCommand.cs
@@ -11,8 +11,11 @@ namespace ElectronNET.CLI.Commands
     {
         public const string COMMAND_NAME = "start";
         public const string COMMAND_DESCRIPTION = "Start your ASP.NET Core Application with Electron, without package it as a single exe. Faster for development.";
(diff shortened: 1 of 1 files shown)


robertmuehsig:
@callumhutchy Sorry for the huge delay. In threory your PR looks good and makes sense, unfortunately there is a merge conflict because of recent changes. 
If you are still interested to see this feature in a future release I'm happy to merge if you could resolve the merge conflict.

callumhutchy:
@robertmuehsig I'll review this at some point over the weekend

gabecook:
This may help. It's what we are using with the current version of ElectronNET.

https://github.com/ElectronNET/Electron.NET/pull/463

GregorBiswanger:
We use #463

PR #313, opened 2019-10-03
Update to .NET Core 3.0

robertmuehsig:
Based on this PR https://github.com/ElectronNET/Electron.NET/pull/300 all packages are updated to .NET Core 3.0 + the build files should point to the correct environment.

Not sure if on the NPM part, but we will see what the CI Systems tells us :)

PR #328, opened 2019-10-23
Converted global npm dependencies to use npx

jimbuck:
These changes remove the need to either A) install tools globally (`typescript` and `electron-builder`) and B) call tools locally via actual path (some `tsc` calls find the bin path first).

This should avoid any differences between platforms but I haven't been able to test it on every platform (just Windows for now). I should be able to run some tests on Linux this weekend.

See [my comment](https://github.com/ElectronNET/Electron.NET/pull/313#issuecomment-545256352) on #313 for more details.


robertmuehsig:
Oh - this looks nice! Thanks!

PR #331, opened 2019-10-28
This should fix #318

### LINKED ISSUE #318: Rename Electronize.exe for better Azure DevOps support
If you change the name of `electronize.exe` to `dotnet-electronize.exe`, when the global tool is installed, you'd be able to use the built-in Azure DevOps .NET Core tasks to call `electronize`, instead of having to use Powershell.

robertmuehsig:


GregorBiswanger:
@robertmuehsig "This branch has conflicts that must be resolved"

Issue #308, opened 2019-09-15
Start process with listen port 8000 error.


MiyamuraMiyako:
5.22.14

.Net Core3.0 Preview9

amd64/windows10

Just simple init project, start it will be throw exception.

Uncaught Exception:
Error: listen EACCES: permission denied 127.0.0.1:8000
    at Server.setupListenHandle [as _listen2] (net.js:1216:19)
    at listenInCluster (net.js:1281:12)
    at GetAddrInfoReqWrap.doListen [as callback] (net.js:1420:7)
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:63:10)

netpoetica:
This issue should be resolved, maybe you are using an older version? It was initially reported here: https://github.com/ElectronNET/Electron.NET/issues/261

And later solved here: https://github.com/ElectronNET/Electron.NET/pull/262

Can you verify that you have these changes? https://github.com/ElectronNET/Electron.NET/pull/262/files

thecodejedi:
I also have this problem but also using 5.22.14 already. I would propose to go for a different solution by limiting the port range

thecodejedi:
#334 pull request for this suggestion

Pichl0r:
The same issue is occuring for me with version 5.30.1

GregorBiswanger:
Fixed with Electron.NET Version 7.30.2

azraelrabbit:
i'm useing ElectronNET.CLI 7.30.2 , but the same error still exist.
![issue-electronnet](https://user-images.githubusercontent.com/553991/70382139-a8226480-1991-11ea-8678-909f025884a1.png)

azraelrabbit:
update, 
i found that my problem just on debugging use visual studio 2019 16.4.0 ,
when i typed `electronize start ` in cmd.exe without visual studio ,that works fine.

thecodejedi:
Maybe cmdine install version is different than the one referenced by the project?


On Mon, Dec 09, 2019 at 5:40am, 大桥 < notifications@github.com [notifications@github.com] > wrote:
update,
i found that my problem just on debugging use visual studio 2019 16.4.0 ,
when i typed electronize start in cmd.exe without visual studio ,that works fine.

nazar322:
Same issue happens on 23.6.1
https://i.imgur.com/JmGPKOu.png
https://i.imgur.com/e3ATeQS.png

nazar322:
[...]

PR #334, opened 2019-11-02
#308 - Start process with listen port 8000 error.

thecodejedi:
excluding the already used port for lookup, so no race condition can happen on the used electron port


thecodejedi:
#308 Still experiencing the problems with Linux (every second time) so it looks like a race condition. Ruling out the port initially solves the problem

PR #351, opened 2019-12-12
added /dotnet-configuration parameter for electronize start command

dmitrya1979:
if your .net core project has several configurations use `/dotnet-configuration` parameter

```
electronize start /dotnet-configuration MacOsRelease
```

### DIFF
--- a/ElectronNET.CLI/Commands/StartElectronCommand.cs
+++ b/ElectronNET.CLI/Commands/StartElectronCommand.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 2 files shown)


GregorBiswanger:
Please set /dotnet-configuration as an optional parameter for `electronze start`.

Maybe better if we can optionally store the configuration in the electron.manifest.json? Also have the possibility to store different electron.manifest.json here. This would simplify the organization for different environments considerably.

As example:
  * Create a additional manifest file: `electronize init /manifest test`
  * Start the app with your additional manifest file: `electronize start /manifest electron.manifest.test.json`
  * Build the app with your additional manifest file: `electronize build /target win /manifest electron.manifest.test.json`.

robertmuehsig:
This seems to be related to https://github.com/ElectronNET/Electron.NET/pull/324

The start and build commands are related and they should support similar arguments. Either someone updates the PR 324. 

I would say we close this PR and if someone is willing to implent it, the PR 324 way would be better IMHO.

PR #376, opened 2020-02-25
added option for custom package.json

Sachin4dotnet:
added option for custom package.json in electronize start comman
similar like manifest was added


GregorBiswanger:
All checks have failed - please fix it - Then a simple copy is not enough for everything else to work. Please have a look at Electron.NET HostHook feature. Here you can create and use your own package.json.

GregorBiswanger:
Electron.NET is not compatible with lower versions of Electron. It all depends on the Electron.NET versioning. As an example the next version will be: 8.31.1 - 8 stands for the supported electron version. The 31 for .NET Core 3.1 - and the 1 for our release from the new Electron version.

Compatibility for an electron / node version can be introduced with the electron-rebuild super package.

Alternatively, you can manage the Electron HostHook manually different package.json files yourself.

PR #402, opened 2020-05-17
Remaining implementation of the NativeTheme API

konstantingross:
'shouldUseHighContrastColors', 'shouldUseInvertedColorScheme', get/set 'themeSource' and 'updated' event added


ElectronNET.API/NativeTheme.cs
```diff
@@ -32,11 +33,114 @@ internal static NativeTheme Instance
...
 
         /// <summary>
-        /// A `Boolean` for if the OS / Chromium currently has a dark mode enabled or is
-        /// being instructed to show a dark-style UI.If you want to modify this value you
-        /// should use `themeSource` below.
+        /// Checks if the new ThemeSource is valid.
+        /// </summary>
+        /// <param name="themeSource">The new ThemeSource to check.</param>
+        /// <returns>True, if is a valid ThemeSource.</returns>
+        internal bool IsValidThemeSource(string themeSource)
```
GregorBiswanger: Can you offer this to an enum? Is type-safe and guides the developer in the flow ...

ElectronNET.API/NativeTheme.cs
```diff
@@ -32,11 +33,114 @@ internal static NativeTheme Instance
...
+        /// <description>Dark Mode: SetThemeSource("dark");</description>
+        /// </item>
+        /// <item>
+        /// <description>Light Mode: SetThemeSource("light");</description>
+        /// </item>
+        /// </list>
+        /// Your application should then always use <see cref="ShouldUseDarkColorsAsync"/> to determine what CSS to apply.
+        /// </summary>
+        /// <param name="themeSource">The new ThemeSource.</param>
+        public void SetThemeSource(string themeSource)
```
GregorBiswanger: Here the same please...

ElectronNET.API/NativeTheme.cs
```diff
@@ -32,11 +33,114 @@ internal static NativeTheme Instance
...
+        {
+            // Check for supported themeSource, otherwise it sets the default
+            if (!IsValidThemeSource(themeSource))
```
GregorBiswanger: No must.. but it is nicer to check for a truth than for an if not equal then... (clean code)

konstantingross: I didn't know if I was allowed to turn everything completely upside down :D With an enum I can delete this method.

PR #405, opened 2020-05-18
Update application exit logic

### LINKED ISSUE #346: OSX Processes not closing
Related to issue #226 and #68 but #68 was never solved and the solution proposed in #226 doesn't work.

* **Version**: 
Electron.NET API version: v7.30.2
Electron.NET CLI version: v7.30.2

.NET CORE: 3.0
NPM: 6.12.0

* **Target**: 
macOS Mojave

When an app on macOS is closed, its child processes still remain active and have to be killed through the activity monitor. At first I thought this was related to an issue in my code but this issue also exists on the demo projects you provide.

Steps to Reproduce:

1. Build latest demo project on macOS
2. Run
3. Close the app

dafergu2:
On macos, applications using ElectronNET.API were not exiting properly.  Specifically, there are two issues contributing to this, both of which are discussed in #346.

1) The all windows closed event was not firing for macos.  This meant that the application developer was not able to automatically close the application when all windows were closed and running on macos.

2) Depending on how the application was triggered to close, the .NET Core kestrel process was not automatically exiting.

This PR fixes both of these issues.  A larger issue, though, is that application behavior for macos is being fixed in the electron main process.  It may make sense to make a larger change in which all of the behavior is handled in the Electron.API and consuming application, thus giving the end user complete control over the behavior without introducing inconsistent behavior between platforms.  For example, rather than then current logic checking the platform in the electron main process, the ElectronNET.API could check and automatically set the default of `Electron.WindowManager.IsQuitOnWindowAllClosed` on macos to false.  That would eliminate custom logic in the electron main process and keeps behavior consistent.

Since this issue is impacting myself and others, I propose to merge this current change and make the additional change at a later time if deemed appropriate.

Fixes #346


dafergu2:
@GregorBiswanger @robertmuehsig Are these build issues expected?

ElectronNET.Host/api/app.ts
```diff
@@ -1,13 +1,20 @@
...
+
+    // By default, quit when all windows are closed
     app.on('window-all-closed', () => {
         // On macOS it is common for applications and their menu bar
         // to stay active until the user quits explicitly with Cmd + Q
-        if (process.platform !== 'darwin' &&
-            isQuitWindowAllClosed) {
+        if (process.platform !== 'darwin' && isQuitWindowAllClosed) {
             app.quit();
+        } else if (!!appWindowAllClosedEventId) {
```
GregorBiswanger: Normally no two exclamation points necessary... reads more complicated... JavaScript already understands what you want :) it only goes through if it has a valid value.. change and test with `else if (appWindowAllClosedEventId) {`

dafergu2: I've gotten into the habit of `!!` simply due to this:

https://brianflove.com/2014-09-02/whats-the-double-exclamation-mark-for-in-javascript/

`var foo = ''` will cause `if (foo) { console.log('foo is true!'); }` to output 'foo is true'.  In reality, I want to treat an empty string as false.  Thus, a `!!` will convert the var to a bool and then take the inverse.  `if (!!foo) { console.log('foo is true!'); }` will not output a value.  This means I don't have to do any extra checking of undefined, null, or empty string.

Let me know if you'd rather me swap it to the single `!`.  Happy to do so.

PR #406, opened 2020-05-19
Added UserAgentFallback to App

Mandrakia:
Added UserAgentFallback to the App object


ElectronNET.API/App.cs
```diff
@@ -1527,6 +1527,34 @@ public void DockSetIcon(string image)
...
+                    BridgeConnector.Socket.On("appGetUserAgentFallbackCompleted", (result) =>
+                    {
+                        BridgeConnector.Socket.Off("appGetUserAgentFallbackCompleted");
+                        taskCompletionSource.SetResult((string)result);
+                    });
+
+                    BridgeConnector.Socket.Emit("appGetUserAgentFallback");
+
+                    return taskCompletionSource.Task;
+                }).Result;
```
konstantingross, line 1551: Hmm .Result can causes [deadlocks](https://blog.stephencleary.com/2014/12/a-tour-of-task-part-6-results.html). The cleanest implementation would be:
public Task<string> GetUserAgentFallback 
public void SetUserAgentFallback

Or what do you guys think?

GregorBiswanger: I think it would be a problem with functions that are called regularly during application runtime. I would leave it at Properties.

GregorBiswanger: I think some application design is important too... otherwise we would only have methods..

GregorBiswanger:
@Mandrakia  Thanks for your contribution. 
Write me your postal address on Gitter and you will get Electron.NET sticker :)

PR #430, opened 2020-06-12
HostHook.CallAsync should use TaskCompletionSource.SetException instead of throwing exception

freosc:
Actually, this pattern should be updated in the rest of the code as well.
Anywhere you use a TaskCompletionSource, you have to make sure the setResult or setException is called, otherwise the calling code will await forever.


GregorBiswanger:

PR #435, opened 2020-06-15
Enabled empty tray icon

writzx:
This allows empty tray icon to be placed without a context menu. Useful for cases when only minimize/close to tray is implemented.

This somehow also fixes menu icon click events in macOS. (probably a thing in base electronjs as well). Not sure about other OS though.


GregorBiswanger:

PR #463, opened 2020-08-07
Add target command line option found in 'build' to 'start' in CLI to …

gabecook:
…support custom targets in development.

Hello! 

Thank you for all of your work on Electron.NET. 

I added the ability to specify the platform target to the start subcommand in the CLI. I needed this because my application is dependent on several x86 C APIs and it will not build in x64, even in debug. 

For reference, the application is Blazor server side.


GregorBiswanger:

Issue #464, opened 2020-08-13
buildall.sh fails on linux - NETSDK1095: Optimizing assemblies not supported for osx/win


duncanawoods:
Here is the full log:

```
Start building Electron.NET dev stack...
Restore & Build API
  Determining projects to restore...
  All projects are up-to-date for restore.
Microsoft (R) Build Engine version 16.6.0+5ff7b0c9e for .NET Core
Copyright (C) Microsoft Corporation. All rights reserved.

  Determining projects to restore...
  All projects are up-to-date for restore.
  ElectronNET.API -> /home/duncan/Projects/forks/Electron.NET/ElectronNET.API/bin/Debug/netcoreapp3.1/ElectronNET.API.dll
  Successfully created package '/home/duncan/Projects/forks/Electron.NET/artifacts/ElectronNET.API.99.0.0.nupkg'.

Build succeeded.
    0 Warning(s)
    0 Error(s)

Time Elapsed 00:00:01.88
Restore & Build CLI
  Determining projects to restore...
  Restored /home/duncan/Projects/forks/Electron.NET/ElectronNET.CLI/ElectronNET.CLI.csproj (in 153 ms).
Microsoft (R) Build Engine version 16.6.0+5ff7b0c9e for .NET Core
Copyright (C) Microsoft Corporation. All rights reserved.

  Determining projects to restore...
  All projects are up-to-date for restore.
  ElectronNET.CLI -> /home/duncan/Projects/forks/Electron.NET/ElectronNET.CLI/bin/Debug/netcoreapp3.1/dotnet-electronize.dll
  Successfully created package '/home/duncan/Projects/forks/Electron.NET/artifacts/ElectronNET.CLI.99.0.0.nupkg'.

Build succeeded.
    0 Warning(s)
    0 Error(s)

Time Elapsed 00:00:01.11
Restore & Build WebApp Demo
  Determining projects to restore...
  All projects are up-to-date for restore.
Microsoft (R) Build Engine version 16.6.0+5ff7b0c9e for .NET Core
Copyright (C) Microsoft Corporation. All rights reserved.

  Determining projects to restore...
  All projects are up-to-date for restore.
  ElectronNET.API -> /home/duncan/Projects/forks/Electron.NET/ElectronNET.API/bin/Debug/netcoreapp3.1/ElectronNET.API.dll
[...]

PR #465, opened 2020-08-13
buildAll.sh linux errors

duncanawoods:
There are two issues:

- the original directory is not restored after it is invoked - I have used pushd/popd
- #464 the win/osx builds fail on linux - I have disabled them if it's linux - I don't know if they work on osx


duncanawoods:
I see that this doesn't build the electron host typescript. I also suggest adding:

```
pushd $dir//ElectronNET.Host
    npm run-script start
popd
```

GregorBiswanger:

PR #478, opened 2020-08-31
Added support for launching the application with a file on MacOS

dlitty:
Added support for launching the application with a file on MacOS by listening for the open-file and open-url events from the operating system, and forwarding them to the Electron.NET app when listeners are added

### DIFF
--- a/ElectronNET.Host/main.js
+++ b/ElectronNET.Host/main.js
@@ -13,6 +13,8 @@ let powerMonitor;
 let splashScreen, hostHook;
 let mainWindowId, nativeTheme;
 let dock;
+let launchFile;
+let launchUrl;
 
 let manifestJsonFileName = 'electron.manifest.json';
 let watchable = false;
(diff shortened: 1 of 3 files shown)


ElectronNET.Host/main.js
```diff
@@ -194,7 +208,35 @@ function startSocketApiBridge(port) {
         browserView = require('./api/browserView')(socket);
         powerMonitor = require('./api/powerMonitor')(socket);
         nativeTheme = require('./api/nativeTheme')(socket);
-        dock = require('./api/dock')(socket);
+		dock = require('./api/dock')(socket);
+		
+		socket.on('register-app-open-file-event', (id) => {
```
danatcofo, line 213: This feels like the wrong place for this addition. There is no other targeted event handling of this nature in main.js.  I would think that this belongs in api/app.ts or somewhere similar.

dlitty: The problem with that is timing.  The electron.js docs explicitly state that those events must be done as early as possible in the startup - see https://www.electronjs.org/docs/api/app#event-open-file-macos.  I had initially tried putting it into app.ts, but by the time it got there, the ready event had already fired, so it was too late.

danatcofo: What about placing it in the app constructor.  That looks like it fires before the calls here.

dlitty: Unfortunately that does not work either.  I tried moving it into the app constructor and it did not work.  I also added some logging using electron-log to show the order of events:

[2020-10-11 11:39:05.673] [info] old location of will-finish-launching handler
[2020-10-11 11:39:06.805] [info] inside constructor
[2020-10-11 11:39:06.815] [info] old location of register handlers

The first line there is where I have the will-finish-launching handler being installed in the original PR.  That is the key one -  if you miss that event, the whole thing doesn't work.  By the time it gets into the constructor, it is already too late.

HaydnDias:
Is it possible to consider similar changes for windows/linux too? Would be extra useful for handling things like oidc callbacks

PR #486, opened 2020-10-06
442: maintain references between socket.io connection events

### LINKED ISSUE #442: socket.io connection event after disconnect event loses all browserWindow references
* **Version**: 9.31.1

* **Target**: Windows 10

When the socket in main.js gets disconnected it deletes all the requires and then it will connect again. When the connection event fires again it reinitializes all of the requires however the windows[] and window variable aren't ever set in browserWindows.js. To fix it the main.js just needs to copy the references of all the windows before it deletes the require cache for browserWindows.js then after the browserWindows.js gets reinitialized/required the main.js needs to pass the windows to it.
Sorry for the bad explanation. I'm terrible at explaining things. If you have any questions or need clarification let me know.

danatcofo:
fixes #442 

remove deletes of modules during disconnect
add delete of hostHook during disconnect
check if modules exist before importing them during connect
move local caches of modules into global scope.


GregorBiswanger:
Thanks for your PR!

Can you please change the var declarations to const or let.
var should no longer be used. Then the function declaration back into an arrow function.

Were you able to test a refresh of the app with your code or that it still runs in the event of a disconnect?

Then unfortunately we have a couple of conflicts that would have to be resolved for a merge.

Thanks!

danatcofo:
I was able to test this locally with a forced connection close.  it reconnected without issue and all references were maintained without breaking logic links across disconnect/connect.

I'll get this updated for your coding style preferences.

GregorBiswanger:
I merged everything, but the Test WebApp only shows the splashscreen. The application itself no longer responds. Could it be now that the events are no longer listening at the right time?

Please need a solution so that we can roll out the next update.

Thank you very much for your commitment

GregorBiswanger:
Possibly delete the ElectronNET.WebApp/obj/host folder, so that you get the current host version.

GregorBiswanger:
Okay, the problem is that the latest Socket.IO version is not supported by our outdated .net library for Socket.IO.. 🤔😔

So it was my mistake, I have to continue to use the old version ..

danatcofo:
yah, easy to lose something in the middle of multiple changes.

Issue #491, opened 2020-10-22
Upgrade to NET 5


realLiangshiwei:
Hi team.
Are there any plans to upgrade to NET 5?

GregorBiswanger:
Hi @realLiangshiwei, definitely!

realLiangshiwei:
https://devblogs.microsoft.com/dotnet/announcing-net-5-0-rc-2/
RC2 is a “go live” release; we are supported using it in production. We can upgrade now.

techmind-git:
I was wondering if we can use it with .net 5 with the current latest version since it was presented like that here: https://www.youtube.com/watch?v=KHSJQ0nELlY. Are there things that might not work since you've got this issue open and marked for a future milestone? I was a bit confused so maybe you can share some more on this. Thank you.

aschaefer94:
Any ETA on this feature? We're wanting to update to 5.0 but can't until we can build our electron package.

scottkuhl:
I've submitted a PR to address the upgrade to .NET and also upgraded TypeScript and all NuGet packages to the latest version.

aschaefer94:
Just for everyone to know - I found that if you are updating your .NET project to .NET 5, as long as you have dotnet 3.1 on your machine (or your build server or whatever you're using) you can still run your project in the current version of Electron.NET.

PR #509, opened 2020-11-21
Upgrade to .NET 5

scottkuhl:
Upgraded projects from .NET 3.1 to .NET 5.0.

Upgraded WebApp project from TypeScript 3.8 to TypeScript 5.0.

Upgraded NuGet packages to the latest versions:
- Microsoft.SourceLink.GitHub 1.0.0-beta2-19367-01 > 1.0.0
- System.Drawing.Common 4.7.0 > 5.0.0
- Microsoft.VisualStudio.Web.CodeGeneration.Design 3.0.0 > 5.0.0

This addresses issue #491

### DIFF
--- a/ElectronNET.API/ElectronNET.API.csproj
+++ b/ElectronNET.API/ElectronNET.API.csproj
@@ -36,12 +36,12 @@ This package contains the API to access the "native" electron API.</Description>
    <FrameworkReference Include="Microsoft.AspNetCore.App" />
 </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-beta2-19367-01">
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0">
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
     </PackageReference>
     <PackageReference Include="SocketIoClientDotNet" Version="1.0.5" />
-    <PackageReference Include="System.Drawing.Common" Version="4.7.0" />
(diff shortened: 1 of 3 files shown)


scottkuhl:
I don't see any testing guidelines on the repo.  I was able to run the web app to test the changes and fire off demo code.

image

Delpire:
Will this change mean that our existing .NET Core 3.1 projects will no longer be able to update versions of Electron .NET?

scottkuhl:
If you were to use the .NET 5 version I think you would need to update your project to .Net 5 as well.  From Core 3.1 that should be possible but of course not everyone will want to do that.

Delpire:
Yeah I am not sure we'll want to do that. .NET Core 3.1 and then .NET 6 are the long term support versions from Microsoft. I would still like to be able to get fixes and new features from Electron .NET without having to upgrade to .NET 5. 

Unless you guys are planning on adding some .NET 5 specific features, is there really any reason to drop .NET Core 3.1 support. Could you not still build and ship both in a NuGet package?

scottkuhl:
That's probably a better question for Gregor to answer.  He is planning out the next milestone release now.

ErikApption:
Fantastic - I could really use this version - when is it going to hit nuget?

GregorBiswanger:
I think the new version will only be compatible with .NET 5. However, the 9.31.x version can still be used. I think otherwise we will get mixed up on two tracks.

arne309:
Is there any ETA on when this will be available on github?

GregorBiswanger:
@arne309 I still have three weeks vacation. Definitely will publish it in the next few weeks.

ErikApption:
It looks like this feature is merged - is there any bug holding from doing a beta release in nuget?

Issue #510, opened 2020-11-21
Upgrade Electron to Version 11


scottkuhl:
Upgrade the version of Electron that Electron.NET is using from 9 to 11.

scottkuhl:
I started a POC here: https://github.com/scottkuhl/Electron.NET/tree/poc/upgrade-to-electron-11

That repo updates all NPM packages to the latest along with Electron to v11.  It looks like there is one breaking change that will affect Electron.NET: https://github.com/electron/electron/pull/23578  The id property has been removed from BrowserView.

I don't see anywhere this code is used on the web app for testing.  And since there are no automated tests, how do you normally go about testing these kinds of changes?

GregorBiswanger:
Our next update will support Electron 11 + .NET 5.

scottkuhl:
@GregorBiswanger 

I have a PR in for .NET 5 and I'm working on the Electron upgrade to 11 in the linked repo above.  Should I just hold off and wait?

GregorBiswanger:
Oh no, that´s perfect! I will definitely prefer your PR... because we only have time for the big update at the beginning of December...

PR #511, opened 2020-11-21
Upgrade Electron from 9 to 11

scottkuhl:
Updates all the NPM packages including Electron from 9 to 11.

This addresses issue #510


scottkuhl:
The web app is up and working on the new version.  BrowserView may need an update to how it handles ID, but the ID referenced in C# does not seem to refer to the id referred to in [breaking changes for Electron 11](https://www.electronjs.org/releases/stable#11.0.0).

GregorBiswanger:
Thanks for your PR!

We will now migrate to Electron 11.1.1

PR #528, opened 2021-01-19
Issue/527

### LINKED ISSUE #500: The 32 bit exe generation code wont run in package manager or cmd console.   a double quote is missing, wasted many days of work due to this simple issue
electronize.exe build /target custom win-x86;win /dotnet-configuration Debug /electron-arch ia32
wont work but 

electronize.exe build /target custom "win-x86;win" /dotnet-configuration Debug /electron-arch ia32
will work

the difference here is  "win-x86;win", double quotes

so i request you to fix the documentation

### LINKED ISSUE #527: MacOs enhanced support
There are at least 2 different MacOs targeted features that are missing which are kind of important to a fully functioning app.  

* Dock.SetMenu() which is currently in a TODO state
* App.Activate listener to listen for the `app.on('activate')` event that indicates a dock click.  the activate event also is fired off for other events but for the purposes of Electron.NET, the dock click is the primary missing feature.  Without this its impossible to know when the dock icon has been clicked and no windows are open.

I'm planning to implement these myself and create a pr for them, but wanted to open up this feature as a handle to gather any other major macOS interactions that are missing.

danatcofo:
This PR contains a couple different things separated out int different commits.

* Extension Method for adding the Electron static class members to the standard MS DI Containers, this is a QOL issue only.
* Documentation fix for #500 as it was right there in my face
* SetMenu completed for the Dock (MacOS)
* On and Once implementations for the App and Tray to cover the plethora of events that are not mapped explicitly in those two modules.
* Adding the `EnableRemoteModule` property to the WebPreferences object.  As of Electron 10, this property defaulted to false and without it exposed you can't use the remote module within a window.

Example for the Dock Menu

```csharp
Electron.Dock.SetMenu(new [] {
   new MenuItem {
      Label = "Dock Menu Item",
      Click = () => {
         // do something
      }
   },
});
```

Example for consuming the `activate` event (MacOs only)

```csharp
Electron.App.On("activate", obj => {
   var hasWindows = (bool)obj;
   // do something
});
```

fix #500
fix #527


GregorBiswanger:

IIiKoBiT:
@GregorBiswanger
@danatcofo 


After setting browserWindowOptions.WebPreferences.EnableRemoteModule = true;
The application will stop starting.

image

There is a process, but the window is not created or shown, I'm not exactly sure.

After the update, the code related to selecting a file from the system stopped working for me, as I understood I need to specify browserWindowOptions.WebPreferences.EnableRemoteModule = true; what would it work. But after that, the window completely stops working.

image

danatcofo:
What version of electron are you using?

Issue #545, opened 2021-03-01
Add support for Electron's app `login` event for proxies that require user:pass authentication


silent-0:
Electron's [login](https://www.electronjs.org/docs/api/app#event-login) event is emitted when the BrowserWindow requires basic username/password authentication (e.g. for a website or a proxy). Electron's default behavior is to suppress the dialog window that shows up for this authentication, so if you want to be able to use a proxy that requires auth, you _have to_ catch this event and handle authentication programmatically.

In Electron.js, here's how I would handle this:
```js
app.on('login', (event, webContents, request, authInfo, callback) => {
    event.preventDefault(); // Prevents the username/password dialog from showing up
    let username = "someUsername";
    let password = "somePassword";
    callback(username, password); // Calls the same authentication request that submitting the dialog form would have called
});
```

As far as I can tell, this functionality doesn't seem to exist in Electron.NET, which means using proxies that require authentication isn't possible.

GregorBiswanger:
🎉🚀 New Electron.NET version 23.6.1 released 🚀🎉

With native Electron 23 and .NET 6 support. Your problem should be fixed here. If you continue to have the problem, please let us know. Please note the correct updating of your API & CLI. Info in the README. Have fun!

PR #561, opened 2021-04-26
Add support for proxies that require basic username/password authentication

nfichter:
See issue #545

Electron has a `login` event that allows automatic handling of basic username/password authentication. Because this event includes a callback function, it's difficult to forward it back to Electron.NET. 

So, this PR provides a workaround by adding optional `Proxy` and `ProxyCredentials` fields to the `BrowserWindowOptions` and `BrowserViewConstructorOptions` classes. If these fields are included, `browserWindow.ts` and `browserView.ts` make use of them in their handlers of the `createBrowserWindow ` and `createBrowserView` events, respectively. These files share access to a global `proxyToCredentialsMap` which maps the proxy (`host:port`) to its credentials (`username:password` format). If only a `Proxy` is supplied, they call the `webContents.session.setProxy` method, providing an alternative to explicitly calling that function in Electron.NET. If both a `Proxy` and its `ProxyCredentials` are supplied, they add an entry to `proxyToCredentialsMap`. Supplying only `ProxyCredentials` has no effect.

Finally, `browserWindow.ts` adds an event listener for the `app`'s `login` event. This listener uses the `authInfo`.`isProxy` field to check if the type of basic authorization requested is for a proxy (i.e. the server returned a [407 status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/407)) — there are cases in which basic authorization is used for restricting access to a site (example [here](http://www.httpwatch.com/httpgallery/authentication/authenticatedimage/default.aspx?0.5489813531171844)) in which case a [401 status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401) is returned instead.

The `authInfo` object also provides information about the host and port of the request (in the case of a 401 status code) or proxy (in the case of a 407 status code). This `login` handler checks if the proxy, `${authInfo.port}:${authInfo.host}`, exists as a key in the global `proxyToCredentialsMap`. If it does, and its corresponding value contains only 1 semicolon (as it should, since credentials should be in `username:password` format), then it splits on the semicolon to retrieve the username and password, and submits the callback function which provides authentication to the proxy server.

----------------------------------------------------

Example usage (in `Startup.cs`):

```cs
        public async void ElectronBootstrap()
        {
            var browserWindow = await Electron.WindowManager.CreateWindowAsync(new BrowserWindowOptions
            {
                Width = 1152,
                Height = 940,
                Show = true,
                Proxy = "host:port", // Replace with your own proxy
                ProxyCredentials = "username:password", // Replace with your own credentials
            }, "");

            var browserView = await Electron.WindowManager.CreateBrowserViewAsync(new BrowserViewConstructorOptions
            {
                WebPreferences = new WebPreferences(){
                    Partition = "test-partition", // Partitioning ensures that the BrowserWindow and its BrowserView don't share a session, in case you want to provide separate proxies to each
                },
                Proxy = "host:port", // Replace with your own proxy
                ProxyCredentials = "username:password", // Replace with your own credentials
            });
            browserWindow.SetBrowserView(browserView);

            var bounds = new Rectangle();
            bounds.X = 26;
            bounds.Y = 500;
            bounds.Width = 1100;
            bounds.Height = 400;
            browserView.Bounds = bounds;

            await browserView.WebContents.LoadURLAsync("https://whatismyipaddress.com");
            await browserWindow.WebContents.LoadURLAsync("https://whatismyipaddress.com");

            browserWindow.Show();
            browserWindow.SetTitle(Configuration["DemoTitleInSettings"]);
        }
```

In this example, I used separate proxies for the browserWindow and the browserView. Here is a screenshot of the result:

Screen Shot 2021-04-26 at 2 33 39 PM

The window on the left is a normal Safari window that contains my actual IP. The window on the right is the Electron window — the top half is the BrowserWindow, and the bottom half contains its embedded BrowserView. Notice that each IP is different because of the proxies supplied to each window.


ppostman:
This is exactly what I need for my project. Is there any ETA when this will be merged to the public NuGet package?

silent-juiced:
> This is exactly what I need for my project. Is there any ETA when this will be merged to the public NuGet package?

@ppostman -- if you need this sooner than when they merge it in (if they ever do), what I did was fork the project, implement this, and publish it to NuGet. My .NET project uses my version of ElectronNET on NuGet, instead of the official version. Follow the "Working with this Repo" section of the README, there's a good video to help you out with how to implement this.

PR #579, opened 2021-06-28
Upgrade to use Electron 12.0.12

### LINKED ISSUE #578: Breaking change in electron-updater caused by using Electron 11
CLI: 11.5.1.0
NuGet: 11.5.1

* **Target**: win-x64

As per [the ticket open for electron-updater](https://github.com/electron-userland/electron-builder/issues/6000) all auto update applications will break. As this appears to be the route the developers of the package are heading, the Electron version used in this package should be updated to at least Electron 12.

Steps to Reproduce:

1. Create a project with auto-updater enabled and properly set up
2. See the application throw stating that `Error: Cannot find module 'fs/promises'`

tub5:
Upgrade the ElectronNET.CLI to use Electron v12.0.12 to prevent crashing when the auto update feature is used.

Resolves #578


theolivenbaum:
Something I noticed while testing: Electron 12 has a breaking change with the default value for ``NodeIntegration`` and ``ContextIsolation``, [this change](https://github.com/theolivenbaum/Electron.NET/commit/8f5a7856d5dadb228f5d834674f35cc137bebeea) fixes the default values (but will be a breaking change for Electron.NET)

danatcofo:
@theolivenbaum if you look at the api script around creating browser windows. it forces nodeIntegration to true regardless of the setting in the c# code.  Only the contextIsolation is pertinent here as towards breaking changes.

theolivenbaum:
I thought so, but I think because it has the DefaultValue attribute, it was not being serialized or had the wrong value. But in any case think it's good to adopt the default values from electron even if it's a breaking change moving forward...

nazar322:
I have this issue as well. Waiting for the next release containing this fix. Thanks!

Issue #571, opened 2021-06-07
Cannot run demo app with latest version.


avalon30:
* **Version**:  electronnet.cli (version '11.5.1') | dotnet (version '5.0.300')

* **Target**: Microsoft Windows [Version 10.0.17763.1971]

Steps to Reproduce:
> Execute in PowerShell (eventually git-bash)
1. dotnet tool install --global ElectronNET.CLI
2. git clone https://github.com/ElectronNET/electron.net-api-demos.git
3. cd electron.net-api-demos
4. electronize start

**Expected result:**
https://github.com/ElectronNET/electron.net-api-demos/raw/master/screenshot.jpg

**Actual result:**
_Hangs on this splash screen_
https://i.imgur.com/ICjFKi8.png

**Log:**
https://pastebin.com/raw/Qtvfsi9q

**Comment:**
Demo app cannot start with latest version including latest dotnet.
This framework seems like an amazing alternative to original Electron.js especially for .Net devs. It needs to be updated to handle latest features and fixes. Currently you're unable to build a project to final exe, so you cannot share your app with the world. Yes, `electronize start` works, but nothing else. Good question to be ask: Is this project holds some future or not anymore?.
Thank you all for your hard work and if possible please to update this project 👍

mgamache:
This is the same issue I have with my project (using .net 5.0 / asp.net MVC). 

Note: _**ConfigureServices**_ is called, but not _**Configure**_ on the asp.net startup class. It just sort of stops.

mbparker:
So is this project dead? I got it a demo project working with 11.5.1 with no problem, but the current version (13.5.1) is doing the exact same thing referenced above. It doesn't work at all.

mbparker:
I saw the same thing when I upgraded to 13.5.1, but I had also upgraded my other dependencies. After reverting, and updating one at a time, I narrowed it down to Autofac 6.2.0. Not sure if this helps anyone else. Autofac 6.0.0 seems to work fine.

GregorBiswanger:
🎉🚀 New Electron.NET version 23.6.1 released 🚀🎉

With native Electron 23 and .NET 6 support. Your problem should be fixed here. If you continue to have the problem, please let us know. Please note the correct updating of your API & CLI. Info in the README. Have fun!

PR #592, opened 2021-08-04
Log errors in the dotnet process

Meberem:
I have just used this slightly modified script to debug why my application wasn't starting correctly/at all.

May assist with #571 as I had identical symptoms


danatcofo:
can I recommend also adding this below?

```javascript
apiProcess.on('close', (code) => console.log(`ASP.NET Core Application: exited with code ${code}`));
```

danatcofo:
by the way, we really need this, I added it already to the fork I use for Prefix, it is very useful in solving issues where the net core app doesn't start successfully or crashes.

GregorBiswanger:
@danatcofo I have my big vacation in two weeks and will have time again for PR ..

PR #595, opened 2021-08-18
Switch to async socket lib

### LINKED ISSUE #431: Get properties that use Task.Run().Result seem to block execution
* **Version**: latest from github 9.31.1 

.net core 3.1 node 12.16.3

* **Target**: win

Steps to Reproduce:

1. In the demo Electron.Net WebApp, in ElectronBootstrap, I added these lines of code
```
Console.WriteLine("set AllowDowngrade");
Electron.AutoUpdater.AllowDowngrade = true;
Console.WriteLine("get AllowDowngrade");
//Code execution blocks on the following line
bool allowDowngrade = Electron.AutoUpdater.AllowDowngrade;
Console.WriteLine("AllowDowngrade: " + allowDowngrade);
```
2. The set property works fine, but I noticed that the get property blocks execution, so the console won't show the last line in the code block above.
Inside the property getter code I see that the code is executed correctly, but the socket.on "autoUpdater-allowDowngrade-get-reply" event is never fired.

This bug occurs for all property getters in the electron.net code that use the Task.Run().Result pattern

(Sidenote: a few seconds after the code execution is blocked, the socket also disconnects)

theolivenbaum:
@GregorBiswanger finally had some time to prepare a PR with my changes:

Quick summary of what changed:
- Switch to a supported, asynchronous socket library (https://github.com/doghappy/socket.io-client-csharp / https://www.nuget.org/packages/SocketIOClient) 
- minor changes to the data model to replace arrays with objects on the response side
- Added method to manually configure ElectronNET without using UseElectron(args) on IWebHostBuilder (something I needed)
- Added new parameter to Start command to run from build output and not from publish output
- Fix deadlock with reading properties twice, caused by the previous handler being removed by a new call to `On(triggerEvent, () => { ... })`

Would probably be good if you do a quick review of the TS changes, I had originally only changed the JS side, then noticed you were using typescript for generating it :)

Closes #431


theolivenbaum:
Added also the following today so error messages are also printed to console by the host process:

```
        apiProcess.stderr.on('data', (data) => {
            console.log(`stderr: ${data.toString()}`);
        });
```

GregorBiswanger (approved):

PR #624, opened 2021-11-21
Support Apple Silicon Natively

bman46:
While the current version works well with Rosetta 2, it is beneficial to be able to compile natively for ARM/Apple Silicon.

One issue that I would like input on is the switch from dotnet 5 to 6 for the CLI tools. Dotnet 6 is the minimum version to be able to target `osx-arm64` and switching the CLI tools to 6 allows for the tools to be compiled and run natively on arm. The changes that I made should still work fine with dotnet 5 and be able to compile for `osx-arm64` as long as dotnet 6 is also installed. It may also make more sense for this upgrade to be a separate pull request and upgrade all of the projects together to dotnet 6.

### DIFF
--- a/ElectronNET.CLI/ElectronNET.CLI.csproj
+++ b/ElectronNET.CLI/ElectronNET.CLI.csproj
@@ -33,7 +33,7 @@
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
-    <WarningLevel>4</WarningLevel>
+    <PlatformTarget>AnyCPU</PlatformTarget>
   </PropertyGroup>
 
   <ItemGroup>
--- a/ElectronNET.CLI/ElectronNET.CLI.csproj
+++ b/ElectronNET.CLI/ElectronNET.CLI.csproj
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
 
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>net5.0</TargetFramework>
     <AssemblyName>dotnet-electronize</AssemblyName>
     <ToolCommandName>electronize</ToolCommandName>
 
--- a/ElectronNET.CLI/ElectronNET.CLI.csproj
+++ b/ElectronNET.CLI/ElectronNET.CLI.csproj
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
 
-    <TargetFramework>net5.0</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
     <AssemblyName>dotnet-electronize</AssemblyName>
     <ToolCommandName>electronize</ToolCommandName>
 
@@ -33,7 +33,7 @@
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
-    <PlatformTarget>AnyCPU</PlatformTarget>
+    <WarningLevel>4</WarningLevel>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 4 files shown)


ElectronNET.CLI/ElectronNET.CLI.csproj
```diff
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
 
-    <TargetFramework>net5.0</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
```
danatcofo: Is moving to net6 required for this?  doing this move would cause a major version upgrade.  I believe that we shouldn't combine the 2 changes into a single pr here.

bman46: Yes and no. In order to compile the tool for native distribution on M1 it is required. However, to compile an application for M1, the tool itself could use .net 5 but .net 6 would also need to be installed. 

I have debated between changing it or not and realize that it is a bit out of the scope of one pr, but I decided to add it and get input on it for now. If i didn't bump the version, at a minimum I would like to add a notification to the user that .net6 is required for arm Mac if they don't have it installed already. Im still looking into a way to do that.

bman46: [https://docs.microsoft.com/en-us/dotnet/core/install/macos#arm-based-macs](https://docs.microsoft.com/en-us/dotnet/core/install/macos#arm-based-macs)

danatcofo: You could run `dotnet --list-sdks` and parse the output in the arm cases and if 6 or greater isn't installed throw an error.

bman46: Thanks for the info, i'll have to give that a try.

bman46: I backed the version down to dotnet 5 and added a check to see if 6 is installed. Thanks for the tip!

ElectronNET.CLI/ElectronNET.CLI.csproj
```diff
@@ -33,7 +33,7 @@
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
-    <PlatformTarget>AnyCPU</PlatformTarget>
+    <WarningLevel>4</WarningLevel>
```
danatcofo: was this intentional?

danatcofo:
This looks good.  I think it helps us later down the road as well.  I'd like to eventually have electronize support both 5 and 6 that we we aren't always forcing the users forward before they are ready.

ElectronNET.CLI/ElectronNET.CLI.csproj
bman46: No, it was an automatic change by Visual Studio.
I reverted it in the latest commit.

schaveyt:
This would be an amazing addition!

bman46:
> This will be an amazing addition! Does this not lay the groundwork for supporting RIDs win-arm64 and linux-x64?

Linux arm and x64 is already supported. However, make sure you set the electron arch to arm64 as it will default to x64. Windows arm is not supported but can be targeted manually. The changes in this PR could certainly be used to support ARM for all Operating Systems, but i'm just targeting MacOS for now as I can test it.

danatcofo:
.NET6 support #636

danatcofo:
As a side note.. even intel based macos target will run on silicone with rosetta installed, tested on my m1 air

`softwareupdate --install-rosetta` prior to installation of the app.

bman46:
> As a side note.. even intel based macos target will run on silicone with rosetta installed, tested on my m1 air

It's a bit faster when run native, even when run with the basic hello world app it seems to be noticeable. I haven't timed it though. I also noted this on the PR comment.
> While the current version works well with Rosetta 2, it is beneficial to be able to compile natively for ARM/Apple Silicon.

robertmuehsig:
I think we should switch to .NET 6 before, because otherwise it would feel "weird". I will look into this PR here and then we come back to this, ok? https://github.com/ElectronNET/Electron.NET/pull/636

bman46:
Thats fine with me @robertmuehsig.

PR #631, opened 2021-12-05
changed the processing of loadUrl at CreateWindowAsync

yannikHoeflich:
**adds if missing**
missing part|default value that gets added
-------------|---------------
protocol     | `http://`
host            |`localhost`
port            | BridgeSettings.WebPort

### DIFF
--- a/ElectronNET.API/WindowManager.cs
+++ b/ElectronNET.API/WindowManager.cs
@@ -6,6 +6,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Runtime.InteropServices;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 
 namespace ElectronNET.API
@@ -79,7 +80,7 @@ namespace ElectronNET.API
         /// </summary>
         /// <param name="loadUrl">The load URL.</param>
         /// <returns></returns>
-        public async Task<BrowserWindow> CreateWindowAsync(string loadUrl = "http://localhost")
+        public async Task<BrowserWindow> CreateWindowAsync(string loadUrl = "http://localhost:{port}")
         {
             return await CreateWindowAsync(new BrowserWindowOptions(), loadUrl);
         }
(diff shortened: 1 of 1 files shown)


ElectronNET.API/WindowManager.cs
```diff
@@ -159,6 +156,22 @@ public Task<BrowserWindow> CreateWindowAsync(BrowserWindowOptions options, strin
             return taskCompletionSource.Task;
         }
 
+        private string ParseLoadUrl(string loadUrl) 
+        {
+            var match = Regex.Match(loadUrl, @"(ht|f)tp(s?)\:\/\/([a-z]+\.)+[a-z]+");
```
danatcofo: why not just use UriBuilder and check the port?

```csharp
var uri = new UriBuilder(loadUrl);

// check host and port on uri and modify as necessary

return uri.ToString();
```

This would be more readable and much easier to grok and understand than doing a regex match.

yannikHoeflich: I just tried that out and yes, it is much easier to read but if you don't have the protocol and host given in the uri string, the UriBuilder gives either a completely wrong response or throws an exception. 
The first two examples give a complete nonsense response and the last one throws an exception
```csharp
Console.WriteLine(ParseLoadUrl("github.com/ElectronNET/Electron.NET/pull/631"));
Console.WriteLine(ParseLoadUrl("ElectronNET/Electron.NET/pull/631"));
Console.WriteLine(ParseLoadUrl("/ElectronNET/Electron.NET/pull/631"));

static string ParseLoadUrl(string loadUrl) {
    var uri = new UriBuilder(loadUrl);
    return uri.ToString();
}
```

Do you have any solution for that?

danatcofo: if the intention here is to go to github.com, then you wouldn't see a string format of `github.com` you'd see `https://github.com`. I think the contrive examples your showing here doesn't really account for the reality of what loadUrl is for.  However to cover the case of receiving a relative url you can make an assumption that any url that doesn't start with http passed in (i.e. any relative url) is intended to be localhost.  Thus the following...

Obviously this is an incomplete solution but it works for the given urls of the test case.

```csharp

Console.WriteLine(ParseLoadUrl("github.com/ElectronNET/Electron.NET/pull/631"));
Console.WriteLine(ParseLoadUrl("ElectronNET/Electron.NET/pull/631"));
Console.WriteLine(ParseLoadUrl("/ElectronNET/Electron.NET/pull/631"));

static string ParseLoadUrl(string loadUrl) {
	Uri BaseUri = new Uri("http://localhost:8000");
	if (Uri.TryCreate(loadUrl, UriKind.Absolute, out var url) || 
		Uri.TryCreate(BaseUri, loadUrl, out url)) {
		Console.WriteLine(url.ToString());
		var uri = new UriBuilder(url.ToString());
		return uri.ToString();
	}
	throw new Exception($"Unable to parse {loadUrl}");
	
}
```

yannikHoeflich: I knew that github.com isn't what loadUrl is really for but I wanted an easy example. I implemented the method now into the WindowManager and commited it, I hope that you like it now better. Thanks for your help.

GregorBiswanger:
@danatcofo What is your opinion on this PR? Does it make sense to take this over?

danatcofo:
I don't see an issue with this PR.  Not sure if its needed but I definitely think it hardens the paths better.

PR #633, opened 2021-12-07
Support DI and Mocking better.

danatcofo:
Convert main static references off of `Electron` into interface implementation and expose the underlying Socket for low level interaction.

This is a utility PR to improve support for testing and also better support standard DI practices.  Additionally included is exposure of the underlying Socket vie the ApplicationSocket class so that we can support more low level socket interactions without having to update the ElectronNET repo directly.

Note: I applied all the comment documentation to the interfaces but did not convert origination to use `<inheritdoc />` as I felt that would be inappropriate without some feedback.


niklasstich:
Hey, 
first of all, thanks very much for your work on this pull request, this makes testing components that use the Electron.NET API so much easier and I hope it can be merged soon!

I still have a question about mocking though: I tried to make a test for a component in my example project which saves a file (see https://github.com/niklasstich/Electronize/blob/master/Electronize/Data/ElectronFileSavingService.cs and https://github.com/niklasstich/Electronize/blob/master/ElectronizeTest/SaveTest.cs for the test). But when I was mocking the `IWindowManager`, I realized that it still returns a `BrowserWindow` instead of an `IBrowserWindow`, which means I couldn't mock the class correctly as I wasn't able to mock a `BrowserWindow` with NSubstitute. Instead, I made a copy of the branch from your pull request and made `IWindowManager` return an `IBrowserWindow` instead (see this diff https://github.com/niklasstich/Electron.NET/commit/8490c2bd1c3c829a6765e7e9bb2c2fbbc1634706). 

Is there something I'm missing about this or does this have to be changed in order to mock the WindowManager correctly?

danatcofo:
@niklasstich Yah, I didn't want to go too far down making EVERYTHING mockable as I didn't know dependencies and changing the return to IBrowserWindow in WindowManager|IWindowManager would be considered a breaking change for the repo.  Not a big deal but still I was trying to avoid breaking changes.

Your correct that it SHOULD be IBrowserWindow but ... breaking change... I'm not opposed per say of doing this but doing it with some feedback from some heavy contributors or users would be welcome.

PR #655, opened 2022-01-21
Support for additional dotnet publish flags.

danatcofo:
Many build scenarios and scripts require certain flags to succeed.  For example `nuget restore` might be required prior to running `electronize build` and support for the `--no-restore` flag in the generated publish command is required.  

This PR adds the capability to add any of the dotnet publish flags to the `electronize build` command by appending the command with `-- [--dotnet-publish-flag(s)]`

Backwards compatibility to keep current builds running is maintained by adding the `--self-contained` flag if not present or overridden by `--self-contained false|--no-self-contained`.  Additionally any other flags that are already built into the design of the command are ignored or stripped.


danatcofo:
I've got a logic error somewhere in here... working on it

Issue #664, opened 2022-02-17
Error on reloading a window after a second window is closed


dlitty:
* **Version**:  latest (13.5.1)

* .NET 5

* **Target**: None; `electronize start` will reproduce

Using the latest Electron.NET API & CLI: if an application has two windows open, then one of them is closed, and you call Reload() on the first one, an error occurs:  

```
A Javascript error occurred in the main process
Uncaught Exception: TypeError: Object has been destroyed
at BrowserWindow.<anonymous>( ... path ... /obj/Host/api/browserWindows.js:205:55 ...
```

Steps to Reproduce:

1.  Create a new .NET 5 ASP.NET web application
2. Follow the instructions in the readme to set up Electron.NET in the project
3. Instead of `Task.Run(async () => await Electron.WindowManager.CreateWindowAsync());`, use:
```
    Task.Run(async () => 
    {
        var mainWindow = await Electron.WindowManager.CreateWindowAsync();
        var secondWindow = await Electron.WindowManager.CreateWindowAsync();
        secondWindow.Close();
        mainWindow.Reload();
    }
```

Alternatively, comment out the Close and Reload lines, and manually close one window and hit Cmd-R on the other.

danatcofo:
Omg reproduction of a bug I couldn't figure out how to reproduce. And so simple! Thank you, you are my hero!

I'm going to fix this on Monday now that I have repo steps.

This bug has been on my annoyatron radar for months but never had the time to dive in depth to isolate and reproduce.  Just happened randomly it seemed like.

dlitty:
Glad I could help!  Never good to have the annoyatron active!

xsoutline:
I have the same problem

```

[Window Title]
Error

[Main Instruction]
A JavaScript error occurred in the main process

[Content]
Uncaught Exception:
TypeError: Object has been destroyed
    at BrowserWindow.<anonymous> (D:\AllProjects\WebApplication2\Browser\bin\Desktop\win-unpacked\resources\app.asar\api\browserWindows.js:190:55)
    at BrowserWindow.emit (node:events:526:28)
    at node:electron/js2c/browser_init:161:14116
    at process.processTicksAndRejections (node:internal/process/task_queues:78:11)
```

I opened window A when the application started with a custom url scheme, and then used the custom url scheme to open the program, triggering the second instance event. I called CreateWindowAsync to create a new window B, then closed window B. When I clicked on the link on in window A,the link has js listening on code, I reported this error.

[...]

PR #668, opened 2022-02-23
Fix for 664, open for comment.

danatcofo:
This is a fix for #664 where the window will sometimes be null on doing a reload.

I am not quite sure it fixes the issue wholesale but it does prevent the javascript error from occurring.  More testing is required and this PR is here for evaluation and comment.  I do not consider it complete at this time.


danatcofo:
note I missreferenced one of the commits.  it was meant to reference #664

PR #670, opened 2022-02-28
Rewrite to signalr

LORDofDOOM:
Hi Guys,

I did a rewrite from SocketIO to signalr because there are a lot of drawbacks if there are 2 "server" run concurrent (SocketIO from nodejs and Kaestrel from dotnet). Besides of that I dislike the fact that one program needs 2 ports ;-)

I also want to worry about only one side in terms of security, IMHO dotnet with Kaestrel contains a lot of useful stuff for authorize and we can make sure that only "verified" commands run though.

Please **DONT** merge this as it is - It's it basically working but have a lot of drawbacks and untested areas (besides of the fact that it completely breaks the SocketIO version), so please let me know if you consider to use this as idea/base and take a review. If there are issues (they are^^) just write me.

Known Problems/Issues:

- We wait 5s for each response from Node -> C# (see SignalrSerializeHelper.cs) , this is a problem if we wait for a Window result (like a MessageBox) - After 5s everything from SignalrSerializeHelper will return null, should be easy to implement.

- IpcMain is not working (committed out multiple lines) - Unsure if this is needed because we're able to use signalr/razor/blazor for this.

- Hybrid Support is forced to be true

- Electron HostHook did not work (also unsure if needed)

- Demos are not converted yet

- SignalrSerializeHelper is a bit messy and could get replaced in net7 with https://github.com/dotnet/aspnetcore/issues/5280

Expect a lot of bugs ;-)

Thank you


ElectronNET.Host/ElectronHostHook/package.json
```diff
@@ -7,12 +7,15 @@
...
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "start": "tsc -p ."
   },
   "keywords": [],
   "author": "Gregor Biswanger",
   "license": "MIT",
   "devDependencies": {
+    "@microsoft/signalr": "^6.0.2",
+    "@types/signalr": "^2.2.37",
```
BrennanConroy: @microsoft/signalr ships with its own type definitions. @types/signalr is for the older ASP.NET SignalR and is not compatible with @microsoft/signalr

LORDofDOOM: @BrennanConroy Thank you, thats correct, I've started a while ago with aspnet/signalr and just updated the packages - It's fixed :-)

GregorBiswanger:
We have already received a PR #595 with a complete switch to another .NET library with Socket.IO support. Socket.IO runs natively in the backend and is better supported under Node.js than SignalR.

We are happy to be open to pros and cons when deciding on SignalR. Currently there would be no reason.

LORDofDOOM:
> We have already received a PR #595 with a complete switch to another .NET library with Socket.IO support. Socket.IO runs natively in the backend and is better supported under Node.js than SignalR.
> 
> We are happy to be open to pros and cons when deciding on SignalR. Currently there would be no reason.

This PR is just to have a discussion :-) 

From my point of view it depends on the use case what is a better fit. In my current project I use Electron.NET as completely optional component/plugin and in that scenario it makes sense to use dotnet as "master", I also use Electron.NET as GUI of a remote application that runs on a webserver for a different project - Both cases are not possible with the current implementation because Electron.NET and the main assembly needs to be on the same system and electron need to be the entry application what is not the case with the signalr implementation.

I also dislike the fact that we need to run 2 independent "server" implementations that use a separate port (dotnet backend, nodejs socket io) and there is no way to have a useful authorization - This problem dont exists with signalr/kaestrel as backend solution and is a show stopper in any real world scenario where security is a factor. Another thing that will make lazy people happy is that you dont need to attach a debugger - you can just debug from VS as you do in any other dotnet application :-)

GregorBiswanger:
The two processes communicating via web sockets is ugly. I'm with you there. With SignalR we would have the same communication as well. I haven't noticed any difference in your implementation.

As far as client application security is concerned, any desktop solution is easily vulnerable. Processes hosted in VMs can also be easily dismantled with a kernel debugger.

But yes, an alternative for Electron.NET would be great. We just haven't found a simple solution to this.

LORDofDOOM:
> The two processes communicating via web sockets is ugly. I'm with you there. With SignalR we would have the same communication as well. I haven't noticed any difference in your implementation.
> 
> As far as client application security is concerned, any desktop solution is easily vulnerable. Processes hosted in VMs can also be easily dismantled with a kernel debugger.
> 
> But yes, an alternative for Electron.NET would be great. We just haven't found a simple solution to this.

Thats true - In terms of communication it's pretty similar, at the end of the day signalr is also a socket connection.

We need to run kaestrel (dotnet) in any scenario (with or without socketio) so the dotnet backend process will always be needed and with signalr it's just a additional endpoint on the same already existing backend and don't request a additional backend (socketio) that's hostet in a different process (electron)

In my PR Electron.NET (UI/Electron itself) is optional and not a base requirement that bootstraps the dotnet entry assembly, it's exactly the way around - dotnet will start the electron process if needed and setup/provide the communication with signalr. It shifts the complete server logic to one process (dotnet).

In terms of security concerns: It's way easier to exploit a application where I just need to send a packet to a socket without any kind of authorize than access the system memory.  ATM every application that can communicate with localhost is able to send packets and use Electron.NET as puppet. Other exploits are for sure also possible but e.g. a memory injection would need at least elevated user rights and another level of knowledge.

With one backend the coder is able to use any level of protection/authorization aspnetcore offers (and that's a lot^^) and don't need to take care about two backend services.

Here a little sumup what the signalr implementation offers:

- Only one backend (dotnet), not 2 (dotnet, socketio)
- Need only one port
- Electron and dotnet are independent processes that are no longer bound to each other
- No need to bootstrap dotnet assembly from electron (even if this should still work)
- Easier debugging (no need to attach debugger)
- Electron is optional and can get started at runtime (electron can get started if it's needed but is not required to run the dotnet assembly)
- Electron can run as client/UI only for a remote service on a webserver but still can communicate with the desktop (useful if you want to push notifications from the server to the client or any other kind of desktop interaction)
- Authorization is possible though aspnetcore (e.g. JWT, cookies, OAuth, IdentityServer)
- Encrypted communication is possible
- With net7 (hopefully) it should be also possible to use signalr like a REST api where we no longer need to use the ugly workaround with SignalrSerializeHelper (what is also the case with socketIO, just different implemented)

peter-bozovic:
This looks like interesting solution to me !

But if it is very much different from current Socket implementation I understand it could take a lot of time to fully test this approach and make it production ready.
I'm eager to get the new release that I hope is around the corner, so if we can get what is already stable with sockets would be great. ;)

And would it be possible to have like multiple different host models ? Users could choose if they wish to use the SocketHost library or the SignalRHost that could be the new one ?

I'm using currently Electron with Blazor Server. Would it use the same SignalR in this scenario or it would have two layers of SignalR within SignalR ? :)

And what about the third model that could work like BlazorWebView that Microsoft is shipping for MAUI and WPF ?
Can we manipulate directly the DOM elements within Electron Web View without having to use intermediate communication system ?

I'm really enjoying so far the use of electron.net and hoping for the update soon !

LORDofDOOM:
> This looks like interesting solution to me !
> 
> But if it is very much different from current Socket implementation I understand it could take a lot of time to fully test this approach and make it production ready. I'm eager to get the new release that I hope is around the corner, so if we can get what is already stable with sockets would be great. ;)
> 
> And would it be possible to have like multiple different host models ? Users could choose if they wish to use the SocketHost library or the SignalRHost that could be the new one ?
> 
> I'm using currently Electron with Blazor Server. Would it use the same SignalR in this scenario or it would have two layers of SignalR within SignalR ? :)
> 
> And what about the third model that could work like BlazorWebView that Microsoft is shipping for MAUI and WPF ? Can we manipulate directly the DOM elements within Electron Web View without having to use intermediate communication system ?
> 
> I'm really enjoying so far the use of electron.net and hoping for the update soon !

Hi,

the solution works for what I need to get done pretty good - I use it already in production: In my productive version the code is a little bit modified from the code in repo because I need to have authorization and multiple security checks.

It should have implemented almost everything from the socketio solution but will break old code (if you have e.g. custom hooks/endpoints in electron JS).

The different hosting models are hard to implement - In the way it's currently written up we need to maintain both JS implementations (signalr works different from the socketio version and all JS files are changed to reflect this)

For Blazor: Thats exactly what I currently use in my productive solution too - Depending on what hosting model you use (Server/Clientside) it is easy to implement and you can directly bind blazor frontend to electron methods with the classes. I currently use a server side blazor hosting with electron.net as client side frontend.

For BlazorWebView: IMHO this is currently out of scope - The way how this is implemented is based on MS edge, not electron. To have a solution like this I suggest to use chromly or CEFSharp :-)

FlorianRappl:
Outdated.

PR #673, opened 2022-03-22
Add ability to set a window's parent using BrowserWindowOptions

### LINKED ISSUE #666: Setting Modal = true in BrowserOptions does not work
* **Version**: 13.5.1

* **Target**:  windows

If you set `BrowserWindowOptions.Modal = true` when creating a new `BrowserWindow`, the created window will not be a modal window even after setting the parent.

Consider this code:

```cs
var parent = Electron.WindowManager.BrowserWindows.First(); // assume a main window exists and this returns it
var child = await Electron.WindowManager.CreateWindowAsync(
            new BrowserWindowOptions
            {
                Show = false,
                Modal = true
            });
child.SetParentWindow(parent);

child.OnReadyToShow += () =>
{
    child.Show();
}

child.OnShow += async () =>
{
    var isModal = await child.IsModalAsync();
}
```

The observed behavior of the child window is that it is NOT a modal window, and you can still interact with the main window (`parent`). The value of `isModal` in the `OnShow` handler is `false`, too.

I haven't investigated this, but it's possible that for modal windows to work the `parent` must be set inside the `BrowserWindowOptions`: https://www.electronjs.org/docs/latest/api/browser-window/#modal-windows

MutatedGamer:
Closes #666

### DIFF
--- a/ElectronNET.API/Entities/BrowserWindowOptions.cs
+++ b/ElectronNET.API/Entities/BrowserWindowOptions.cs
@@ -154,14 +154,14 @@ namespace ElectronNET.API.Entities
         public bool Frame { get; set; } = true;
 
         /// <summary>
-        /// Whether this is a modal window. This only works when the window is a child
-        /// window.Default is false.
+        /// Whether this is a modal window. This only works when <see cref="Parent"/> is
+        /// also specified. Default is false.
         /// </summary>
         public bool Modal { get; set; }
 
         /// <summary>
         /// Whether the web view accepts a single mouse-down event that simultaneously
-        /// activates the window.Default is false.
+        /// activates the window. Default is false.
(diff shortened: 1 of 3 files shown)


MutatedGamer:
I initially only edited `browserWindows.ts` and ran `tsc`, but it resulted in a lot of changes to `js` and `js.map` files. I decided to make the necessary changes to `browserWindows.js` manually for this PR - let me know if you'd like me to run `tsc` instead.

arun04cbe:
Hi @GregorBiswanger @MutatedGamer 
The code for making the popups modal has been merged but when I check in the repo, it seems the code is not reflected.
Can you kindly check this to get the modal popups working?
(FYI - This has not been reflected in the latest Electron.NET version)

PR #706, opened 2022-06-29
fixes unimplimented interfaces / updates socket client

bknoxphysna:
# What?

I am needing to run ElectronNET on apple silicon and the old build doesn't seem to work.

For some reason the latest on master is broken, this implements missing properties in the interface and replaces the SocketIO client that was being used.

# Details
- Add missing methods and getters from `ElectonNet.API/App.cs`
- Add missing methods and getters from `ElectonNet.API/ApplicationSocket.cs`
- Add missing methods and getters from `ElectonNet.API/AutoUpdater.cs`
- Update the implementation of the SocketIO client


bknoxphysna:
@GregorBiswanger It seems like the latest update for net6.0 doesn't build correctly (in the ElectronNET.API project at least)

In this PR I got it closer to working, but I'm wondering if you would mind helping me finish this up.

I am running this on an m1, I fixed the socketIO implementation (I think), which cause the electron app to simply not finish loading.

But now it doesn't find the javascript files, the window that is created just ends up having a blank window with this error:

`chromewebdata/:1 Not allowed to load local resource: file:///`

PR #709, opened 2022-07-09
Cleaned code and switched to .NET Framework 4.8

onionware-github:
Cleaned some code, changed the .NET Framework version from 4.0 to 4.8 (through Visual Studio).

PS: Directory.Create() only creates a folder if it doesn't already exist. If statements with Directory.Exist are therefore not necessary if a folder is to be created if it does not already exist. I did not find tests for Electron.NET. Please verify that it works with .NET Framework 4.8 as well.


GregorBiswanger:
I merged a big PR from @theolivenbaum - please compare with the new source on our repo.

Issue #735, opened 2023-01-18
ShowOpenDialogAsync escapes URL characters


cosmo0:
* **Version**: 13.5.1

Steps to Reproduce:

1. browse for a folder using ShowOpenDialogAsync
2. select a folder using a "+" in its name
=> the "+" disappears

ravi-devitpl:
I am on v.23.6.1 and still facing the issue with ShowOpenDialogAsync.
1. Browse for a folder using ShowOpenDialogAsync to choose a file
2. For example: C:\Program Files\Notepad++\notepad++.exe
3. It returns: "C:\Program Files\Notepad  \notepad  .exe"
Any help or temporary fix to this issue?

Thanks!

FlorianRappl:
No longerOutdated - use `ElectronNET.Core` and `ElectronNET.Core.AspNet`.

See [Wiki / What's New](https://github.com/ElectronNET/Electron.NET/wiki/What's-New).

PR #736, opened 2023-01-18
Fixes issue #735 - paths are URL escaped

cosmo0:
When selecting a path with special characters, they might get removed by the folder selection C#-side (see issue #735).

They are URL-decoded inside the `ShowOpenDialogAsync` method, which doesn't make sense: the decoding, if applicable, should be the responsibility of the JS side.

It's clearly not mandatory because other "dialog" methods, such as `ShowSaveDialogAsync`, don't use this `UrlDecode` and work fine.

This PR removes the URL decoding.


FlorianRappl (approved):
LGTM!

:+1:

FlorianRappl:
Sorry unfortunately the commit has been based on the old `master` and would conflict with the new `main`.

So either re-create with `main` as new base or wait for us to roll this out. Sorry for the inconvenience.

Issue #754, opened 2023-04-02
WebPrefrences.ZoomFactor is int instead of double


Yuvix25:
* **Electron.NET Version**: 23.6.1
* **.NET Version**: 7.0.202
* **Node.js Version**: 18.12.1
* **Target**: Windows

I want to make my app resolution fixed to 1920x1080 no matter what's the screen resolution, while still keeping the app full screen. To achieve that, I tried the following:
```cs
double factor = (await Electron.Screen.GetPrimaryDisplayAsync()).WorkAreaSize.Width / 1920.0;
var window = await Electron.WindowManager.CreateWindowAsync(new BrowserWindowOptions()
{
    WebPreferences = new WebPreferences()
    {
        ZoomFactor = factor,
    },
    Resizable = false,
    Fullscreen = true,
    Minimizable = false,
    Movable = false,
    Frame = false,
    Transparent = true,
    BackgroundColor = "#00000000",
});
```
However, this did not work, as `ZoomFactor` is expecting an int, not a double, unlike the [original implementation in Electron.JS](https://www.electronjs.org/docs/latest/api/web-frame#webframesetzoomfactorfactor) where it is indeed a double.

Can this be fixed?
(From a quick look at the source code, it seems as the only change necessary is `public int ZoomFactor { get; set; }` to `public double ZoomFactor { get; set; }`, but I'm not entirely sure, so I prefer not posting this as a PR)

FlorianRappl:
Yes, I think this can be fixed and that the fix would be indeed as simple as the suggestion. Do you want to make a PR on this? Thanks!

Yuvix25:
Sure, I'll make one!

Yuvix25:
Also, I used ZoomFactor 2 for a few tests, but now when I removed it, it still is set to that. I tried deleting obj and bin folders, setting ZoomFactor to 1, removing ZoomFactor completely, nothing works, it stays zoomed no matter what. Any idea what can cause this?

GregorBiswanger:
Will be taken into account in the **next update release**. 

Thank you @Yuvix25 for your PR.

PR #755, opened 2023-04-02
Change WebPreferences.ZoomFactor from int to double

Yuvix25:
In the original Electron.js, ZoomFactor is a double, not an int.
See #754 for more info.


FlorianRappl (approved):
LGTM - your thoughts @GregorBiswanger ?

GregorBiswanger (approved):

PR #758, opened 2023-04-03
Fixes issue #735 - paths are URL escaped

cosmo0:
When selecting a path with special characters, they might get removed by the folder selection C#-side (see issue https://github.com/ElectronNET/Electron.NET/issues/735).

They are URL-decoded inside the ShowOpenDialogAsync method, which doesn't make sense: the decoding, if applicable, should be the responsibility of the JS side.

It's clearly not mandatory because other "dialog" methods, such as ShowSaveDialogAsync, don't use this UrlDecode and work fine.

This PR removes the URL decoding.

Re-creation of PR #736


FlorianRappl (approved):
LGTM!

cc @GregorBiswanger

GregorBiswanger (approved):

Issue #411, opened 2020-05-21
ElectronNET.API.Entities.WebPreferences.ContextIsolation [DefaultValue(true)]


freosc:
I tested using Electron.net API and CLI 8.31.2
ElectronJS lists the WebPreferences.ContextIsolation default value as false ([https://www.electronjs.org/docs/api/browser-window]).

ElectronNet API decorates the WebPreferences.ContextIsolation property with [DefaultValue(true)] which results in not setting this value in WindowManager.CreateWindowAsync when serializing "options" to a Json object.

This is my ElectronBootstrap code:
```
BrowserWindowOptions options = new BrowserWindowOptions() {
                Kiosk = true,
                AutoHideMenuBar = true,
                TitleBarStyle = TitleBarStyle.hidden,
                WebPreferences = new WebPreferences {
                    NodeIntegration = false,
                    ContextIsolation = true, // protect against prototype pollution
                    Preload = System.IO.Path.Join(AppDomain.CurrentDomain.BaseDirectory, "preload.js") 
                }
            };
```
doing so, results in this error in the browser:

> init.ts:207 Error: contextBridge API can only be used when contextIsolation is enabled
>     at checkContextIsolationEnabled (context-bridge.ts:7)
>     at Object.exposeInMainWorld (context-bridge.ts:12)

When I remove the [DefaultValue(true)] decorator of the ContextIsolation property in the ElectronNet.Api, it seems to work as expected.

GregorBiswanger:
Fixed with the Electron.NET 8.31.3

PR #785, opened 2023-06-20
Update WebPreferences's ContextIsolation default value to true

Yuvix25:
I've encountered the following error: "Error: contextBridge API can only be used when contextIsolation is enabled", even though  I did enable `ContextIsolation`.

In #411, the same issue was reported, and the problem was traced down to the fact that the default value of `ContextIsolation` was set to the wrong value, which made that attribute un-editable.

I then found out that the default value of `ContextIsolation` was changed from false to true since Electron 12, as noted [here](https://www.electronjs.org/docs/latest/tutorial/context-isolation#what-is-it), so I fixed it.

### DIFF
--- a/src/ElectronNET.API/Entities/WebPreferences.cs
+++ b/src/ElectronNET.API/Entities/WebPreferences.cs
@@ -173,7 +173,7 @@ namespace ElectronNET.API.Entities
 
         /// <summary>
         /// Whether to run Electron APIs and the specified preload script in a separate
-        /// JavaScript context. Defaults to false. The context that the preload script runs
+        /// JavaScript context. Defaults to true. The context that the preload script runs
         /// in will still have full access to the document and window globals but it will
         /// use its own set of JavaScript builtins (Array, Object, JSON, etc.) and will be
         /// isolated from any changes made to the global environment by the loaded page.The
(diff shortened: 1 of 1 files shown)


GregorBiswanger:
Thanks for your PR! Have you checked whether the IPC communication is still working afterwards?

Yuvix25:
Not yet. Watching the "Contributing Getting Started" video as we speak.
Will update once I do 👍🏻

Yuvix25:
@GregorBiswanger I compiled it on my side, with the default value now being set to true, but I still get that error for some reason. Any idea what could be the cause?

Yuvix25:
This is the error I'm receiving:
```
node:electron/js2c/renderer_init:2 Unable to load preload script: C:\Users\[username]\[path-to-project]\obj\Host\bin\wwwroot\js\utils.js
(anonymous) @ node:electron/js2c/renderer_init:2
node:electron/js2c/renderer_init:2 Error: contextBridge API can only be used when contextIsolation is enabled
    at checkContextIsolationEnabled (node:electron/js2c/renderer_init:2:5653)
    at Object.exposeInMainWorld (node:electron/js2c/renderer_init:2:5764)
    at Object.<anonymous> (C:\Users\[username]\[path-to-project]\obj\Host\bin\wwwroot\js\utils.js:441:15)
    at Object.<anonymous> (C:\Users\[username]\[path-to-project]\obj\Host\bin\wwwroot\js\utils.js:457:3)
    at Module._compile (node:internal/modules/cjs/loader:1174:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1229:10)
    at Module.load (node:internal/modules/cjs/loader:1044:32)
    at Module._load (node:internal/modules/cjs/loader:885:12)
    at f._load (node:electron/js2c/asar_bundle:2:13330)
    at o._load (node:electron/js2c/renderer_init:2:3109)
```

Also, they did change the default value to `true` now. See the link I posted in the original comment.

Yuvix25:
Maybe I need to remove `<PackageReference Include="ElectronNET.API" Version="23.6.1" />` from the `[project-name].csproj` file? Or perhaps replace it with something local?

Yuvix25:
@GregorBiswanger any idea?

PR #801, opened 2023-08-25
Implicit Conversions to System.Drawing structs

ds5678:
This pull request adds implicit conversions from the following types:
* `Point`
* `Rectangle`
* `Size`

### DIFF
--- a/src/ElectronNET.API/Entities/Point.cs
+++ b/src/ElectronNET.API/Entities/Point.cs
@@ -20,5 +20,14 @@
         /// The y.
         /// </value>
         public int Y { get; set; }
+
+        /// <summary>
+        /// Convert this <see cref="Point"/> to <see cref="System.Drawing.Point"/>.
+        /// </summary>
+        /// <param name="point">The point.</param>
+        public static implicit operator System.Drawing.Point(Point point)
+        {
+            return new System.Drawing.Point(point.X, point.Y);
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


FlorianRappl:
Why implicit? Explicit I would like a bit more, but maybe there is a good use-case for implicit.

ds5678:
It should be implicit because it's allocation free, computationally efficient, and conceptually the same. Making it explicit would just increase code verbosity.

FlorianRappl:
> allocation free, computationally efficient

Implicit and explicit are actually the same. It's just that one thing is done implicitly by the compiler and the other one must be done explicitly. There is no difference regarding required allocations or computational efficiency. So no - those are not good reasons.

Where I'm with you is that implicit is / should be used for casting to a type that is conceptually the same / does not lose information. So this indeed would be in favor of an implicit cast.

All in all I'm on the fence here - I'd like to hear more opinions on this before we settle for one way or the other.

ds5678:
> > allocation free, computationally efficient
> 
> Implicit and explicit are actually the same. It's just that one thing is done implicitly by the compiler and the other one must be done explicitly. There is no difference regarding required allocations or computational efficiency. So no - those are not good reasons.

No, I meant that my implementations for the cast methods were allocation free and computationally efficient. Explicit casts are recommended when the conversions are more costly, as a warning to developers.

FlorianRappl (approved):
I have not heard any (negative) feedback on this one. We should merge it.

PR #822, opened 2023-11-13
Configurable splash screen size

NickRimmer:
Hey hey everyone,

I would like to suggest you tiny changes. I found an issue with splash screen size, somehow it is always less than picture size and blur the image a lot. 

So with settings of custom size I can set splash screen size bigger on 16 pixels and it perfectly in size now for me. I didn't investigate the issue with window size, as it can be related to many things. But configurable  splash screen size is safe and quick solution (;


FlorianRappl (approved):
It's a good one. LGTM!

PR #828, opened 2024-01-10
Update Display.cs to current spec

Yuvix25:
As you can see [here](https://www.electronjs.org/docs/latest/api/structures/display), the current Display.cs file in Electron.NET is not up to date. I've added all the missing fields (except for those which show up in the docs, but do not actually appear in the object we fetch from electron), and checked locally to see that it works.

One thing to note is that I did not change the `Id` value from `string` to `int`, because that caused the `Electron.Screen.GetAllDisplaysAsync()` method to hang forever, and changing it also means breaking backwards compatibility.

### DIFF
--- a/src/ElectronNET.API/Entities/Display.cs
+++ b/src/ElectronNET.API/Entities/Display.cs
@@ -6,17 +6,52 @@
     public class Display
     {
         /// <summary>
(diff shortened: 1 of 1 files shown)


FlorianRappl (approved):
Thanks!

PR #833, opened 2024-02-14
Fix huge memory waste in IpcMain.Once [IMPORTANT]

Yuvix25:
My app uses `IpcMain.Once` to wait for an "all good, next frame" message from the front end. It's updating data at rates well over 60fps, so I'm calling `IpcMain.Once` a lot. I noticed that since I started doing this, my memory usage went waaay up (from 40-50Mb to about 800Mb), so I started investigating memory dumps, and indeed, I found this object:
image
Which I later traced to be the object holding all of the socket listeners.

To fix this issue, all I had to do was replace the `Socket.On` in `IpcMain.Once` with `Socket.Once<object>`, and my memory usage went back down to 40-50Mb.


FlorianRappl (approved):
LGTM!

Issue #837, opened 2024-03-08
Electronize is not running, if only .NET 8 is installed


HeikeHofmann:
We are using the latest nuget packages V23.6.2.

When we start 'electronize build ...', we get this error:

_You must install or update .NET to run this application.
Architecture: x64
Framework: 'Microsoft.NETCore.App', version '6.0.0' (x64)
.NET location: C:\Program Files\dotnet
The following frameworks were found:
  8.0.2 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]_

.NET 6 will be obsolete end of this year and .NET 8 was released last november.
when will be electron migrated to .NET 8 ?

FlorianRappl:
I removed the bug label as this is not a bug. Applications are shipped with a specific framework dependency.

A newer version might / will be released for .NET 8, but right now we are still at .NET 6. As mentioned from your side the official support is still until later this year - so its totally fine.

If you want to fasten up development speed consider contributing to the project either via PRs or financially. Thanks for your support!

ElsterSIGNALISfanclub:
> A newer version might / will be released for .NET 8, but right now we are still at .NET 6. As mentioned from your side the official support is still until later this year - so its totally fine.

Coming back in 2025 to report that this exact error still occurs, and .NET 6.0 has been moved to end of life and is no longer being maintained. Is there a plan or a workaround to get this problem addressed?

felipeleite:
In that case is no longer MacOS compatible anymore. =(

FlorianRappl:
Outdated - use `ElectronNET.Core` and `ElectronNET.Core.AspNet`.

See [Wiki / What's New](https://github.com/ElectronNET/Electron.NET/wiki/What's-New).

PR #846, opened 2024-04-25
Upgrade to .NET 8.0

Lemon73-Computing:


FlorianRappl:
Quick question: What does work by having this changed? Right now you can use Electron.NET with .NET 6 onwards; this change would make it only compatible to .NET 8 onwards. I don't see any .NET 8 API being used, so I wonder - is that really necessary?

Lemon73-Computing:
Now, `electronize` only runs on .net 6. However .net 6 is close to EoS. We can use `electronize` on new version (.net 8) project by my changed.

FlorianRappl:
> Now, `electronize` only runs on .net 6. However .net 6 is close to EoS. We can use `electronize` on new version (.net 8) project by my changed.
> 
> (Relate issue: #837)

Yeah I agree the CLI should work on more targets. Potentially making the CLI independent of .NET (i.e., self-contained) would just solve this. This way we could go for .NET 8, but as its deployed as a self-contained CLI it would also work on any (older - or more recent) platform - escaping the need to install a fixed version of .NET.

I would say let's merge this and let's improve the CLI in the future.

FlorianRappl (approved):

Lemon73-Computing:
Added:
```xml
<PublishSelfContained>true</PublishSelfContained>
```

PR #847, opened 2024-04-29
Add /no-restore and SelfContained property support in CLI

gorsheninmv:
`/no-restore`
It's that MSBuild doesn't have property to control restore policies. These command are implemented in dotnet cli but not in MSBuild.
/no-restore option can be useful for flexlible control in CI\CD pipelines.

`SelfContained`
Optional --self-contained arg will be useful to reduce package size if the required .net runtime is expected to be installed.

The new options maintain backward compatibilty.


FlorianRappl:
LGTM! What's missing to make this PR viable?

gorsheninmv:
> LGTM! What's missing to make this PR viable?

Now the PR is ready for review :)

PR #849, opened 2024-05-01
Get back net6 support

gorsheninmv:
.NET6 support is valid until [November 12, 2024](https://dotnet.microsoft.com/en-us/platform/support/policy), more than half a year, I'm not sure this is a good idea to abandon .NET6 support in Electron.NET.

I would like to suggest making the project multitarget and support both LTS versions (.NET6 and .NET8)

dotnet tool doesn't support self-contaied tools (see https://github.com/dotnet/sdk/issues/9565), so that `<PublishSelfContained>true</PublishSelfContained>` can be useful only for binaries but not for dotnet tool so I decided to remove the line

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,7 @@ jobs:
       uses: actions/setup-dotnet@v4
       with:
         dotnet-version: |
+          6.0.x
           8.0.x
 
     - name: Build
@@ -32,6 +33,7 @@ jobs:
       uses: actions/setup-dotnet@v4
       with:
         dotnet-version: |
+          6.0.x
           8.0.x
 
     - name: Build
--- a/nuke/Build.cs
+++ b/nuke/Build.cs
@@ -71,6 +71,8 @@ class Build : NukeBuild
         }        
     }
 
+    string Framework => Solution.GetProject(DemoTargetLibName).GetProperty("TargetFramework");
+
     protected override void OnBuildInitialized()
     {
         var parser = new ReleaseNotesParser();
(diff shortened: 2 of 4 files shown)


FlorianRappl (commented):
In general the multi target approach works for me - just not sure about the execution in there.

What we'd need is that the published NuGet package contains both frameworks; still the dotnet commands are run with an explicit framework derived from the ElectronNET.WebApp project.

Just wondering if this is achieved here; maybe only the contained tools is then .NET 8, but the libs contains both.

gorsheninmv:
```
The structure of CLI dotnet tool package I got after executing `dotnet pack`
.
|   ElectronNET.CLI.nuspec
|   output.txt
|   PackageIcon.png
|   [Content_Types].xml
|   
+---package
|   \---services
|       \---metadata
|           \---core-properties
|                   553c89be248c4da09d1469f93f9abb2f.psmdcp
|                   
+---tools
|   +---net6.0
|   |   \---any
|   |           dotnet-electronize.deps.json
|   |           dotnet-electronize.dll
|   |           dotnet-electronize.pdb
|   |           dotnet-electronize.runtimeconfig.json
|   |           DotnetToolSettings.xml
|   |           
|   \---net8.0
|       \---any
|               dotnet-electronize.deps.json
|               dotnet-electronize.dll
|               dotnet-electronize.pdb
|               dotnet-electronize.runtimeconfig.json
|               DotnetToolSettings.xml
|               
\---_rels
        .rels
```
The structure of API package I got after executing `dotnet pack`
```
.
|   ElectronNET.API.nuspec
|   output.txt
|   PackageIcon.png
|   [Content_Types].xml
|   
+---lib
|   +---net6.0
|   |       ElectronNET.API.dll
|   |       ElectronNET.API.xml
|   |       
|   \---net8.0
|           ElectronNET.API.dll
|           ElectronNET.API.xml
|           
+---package
|   \---services
|       \---metadata
|           \---core-properties
|                   3528134964fe4b75b8c25a49aef2d054.psmdcp
|                   
\---_rels
        .rels
```

I double checked if the packages worked correctly with my .NET6 app and .NET8 demo webapp. 
Let me know if I should check something else.

FlorianRappl:
No structure looks fine to me!

Let's keep it like that - thanks for checking!

FlorianRappl (approved):

Issue #744, opened 2023-03-23
🎉🚀 Electron.NET Reloaded: The Adventure Begins! 🚀🎉


GregorBiswanger:
Hey Electron.NET community! 

You might have heard the whispers, but it's time to make it official: "**Electron.NET is dead, long live Electron.NET!**" After a brief hiatus, we're back, and the future looks brighter than ever.

You may have noticed that there haven't been any updates lately - this was mainly due to me being the only active member in the project, while my partner wasn't interested in further development. Add to that numerous pull requests and suboptimal communication, which shook the project's foundations. As a father of three, my private life has also taken up a lot of my time.

But now, everything has changed! 🎆 I'm thrilled to announce that **Florian Rappl** has joined the team, and we're working together with renewed energy. We've updated the project to the latest functional state, with a particular focus on the stability of Websockets. With a brand-new library and a smart architecture, we're now more flexible and future-proof than ever before!

The highlights of our upcoming updates? .NET 6 support, native Electron 20 support, and an impressive stability and performance that shatters all expectations! 💥 But that's just the beginning - we've got many more brilliant features up our sleeves that will make Electron.NET even better. Stay tuned!

And hey, we're only human and appreciate your support! 😉 If you'd like to support us with [donations](https://donorbox.org/electron-net), it would help us show our faces around our better halves. 😄

Get ready for the big update coming in the next few days! 🌟 Together, let's embark on an exciting Electron.NET adventure - stay with us!

Your Electron.NET Team

GregorBiswanger:
Electron.NET 23.6.1 has shipped 🚢

New in this release:
⚡️  Upgraded to native Electron 23
⚡️  Upgraded to .NET 6
🎁  New .NET Socket API (Faster and more stable)
🛠️  Bug fixes and improvements

vmeganathan81:
Thanks, this is great news.

xiaoxinxing12:
谢谢，这是个好消息。

kylethomas320:
New release seems very good so far! 
Startup time has increased by like 10 seconds, which is a huge deal.
Socket communication seems stable for now, although I had to make some minor tweaks for it to work compared to the previous version.

Keep up the good work!

GregorBiswanger:
[...]

PR #856, opened 2024-06-12
Update to Electron 31

jibe-dev:
Just like the update to version 30, it does not update any APIs.


RDI-Blake:
Is this project still being maintained? It seems like this and the develop branch is necessary for LTS. Appreciate the effort.

FlorianRappl:
I'd say it's maintained up to the point of "is maintenance relevant". Apparently everyone seems to be good at where this project is - so we'll leave it there. I had some plans (described in #744) to move the project forward, but as seen in the thread there does not seem to be any interest.

PR #895, opened 2025-10-15
Add documentation

softworkz:
I have manually reviewed and revised everything excepting the pages under "API Reference". I also have no intentions to do so - I mean, it's better than before (=nothing) and when somebody finds errors, they can submit a PR to fix it.

The current result can be viewed here: https://github.com/softworkz/ElectronNET/wiki

The included CI for updating the WIki pages is restricted to the branches main and electronnet_core  (mine), so that it won't start overwriting the Wiki before it's in the main branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,9 +41,9 @@ Also you should have installed:
 - Install the following two nuget packages:
 
 ```ps1
-PM> Install-Package ElectronNET.Core
+dotnet add package ElectronNET.Core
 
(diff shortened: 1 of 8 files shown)


softworkz:
Thanks. That's pretty much it, now.
I tried to get the maximum out from very little time, so there are surely some rough edges here and there at the side of documentation. 
For the code, I think that it's all rather solid. 
Have you had a chance to try it - any issues encountered?

Issue #903, opened 2025-11-04
.NET code no longer compiles after PR merge.


agracio:
https://github.com/ElectronNET/Electron.NET/pull/893 was merged into main however it appears something went wrong and the code in main no longer compiles.

FlorianRappl:
Yes, the `Process` class has been updated and brought back. It was one of the classes missing in the referenced origin.

agracio:
@FlorianRappl not sure why you closed it as completed, current code in main does not compile. Is there a pending PR to fix it?

I performed fresh checkouts on both `main` and `develop` branches just in case there were some merge issues.

FlorianRappl:
1. `main` is the release branch. There is no release for the 0.1.0 yet. Right now its a [preview](https://github.com/ElectronNET/Electron.NET/releases/tag/0.1.0-pre.139).
2. `develop` works - if in doubt look at the [action](https://github.com/ElectronNET/Electron.NET/actions/runs/19075476605)

In general build failures are not for the issue tracker (esp. labelled as "bugs", which are issues with the lib - a build failure does not trigger a build). So a bit odd to open a bug for it - but at least it acted as a reminder.

agracio:
1. This is my mistake as I though that final changes have been merged from develop to main.
2. It appears there is a fresh commit to `develop` - it was failing when I opened this issue.

Yes a 'bug' might have been a mistake but as for issue tracker it does make sense to open an issue if there is a problem with the code. Thank you for taking care of the compile error.

FlorianRappl:
> It appears there is a fresh commit to develop - it was failing when I opened this issue.

Yes I know. But when I closed it the build was running again. (2) was a response to your last comment, not the OP.

agracio:
What is the current status of `develop` branch, I mentioned in another issue that the code can benefit from refactoring shaving 1.5k-2k lines of code, but do not want to make changes while the branch is still unstable should I wait for a merge to main first?

FlorianRappl:
You can always make a PR.

Target for any PRs should be `develop`.

agracio:
Unfortunately `develop` branch is still broken. Both test projects ElectronNET.ConsoleApp and ElectronNET.WenApp have incorrect nuget dependencies. 
This results in solution build errors and inability to launch local test app.

I was able to resolve issue locally by updating ElectronNET.Core and ElectronNET.Core.AspNet nuget package references to 0.0.18.

[...]

PR #904, opened 2025-11-04
nuget reference fix and minor cshtml correction

agracio:
As pointed out in https://github.com/ElectronNET/Electron.NET/issues/903#issuecomment-3487763920 current `develop` branch does not build locally and is unable to launch demo project.

- Fixed incorrect nuget references to point to latest pre-release version.
- Minor fix to cshtml `<` tag in demo project that was not displaying correctly when launched and highlighted as error in some IDEs such as Rider.

### DIFF
--- a/src/ElectronNET.WebApp/Views/About/Index.cshtml
+++ b/src/ElectronNET.WebApp/Views/About/Index.cshtml
@@ -20,7 +20,7 @@
 {
     return WebHost.CreateDefaultBuilder(args)
         .UseElectron(args)
-        .UseStartup&lt;Startup>()
+        .UseStartup&lt;Startup&gt;()
         .Build();
 }</code>
                     </pre>
--- a/src/ElectronNET.WebApp/Views/HostHook/Index.cshtml
+++ b/src/ElectronNET.WebApp/Views/HostHook/Index.cshtml
@@ -42,7 +42,7 @@
     };
     var folderPath = await Electron.Dialog.ShowOpenDialogAsync(mainWindow, options);
 
-    var resultFromTypeScript = await Electron.HostHook.CallAsync&lt;string>("create-excel-file", folderPath);
+    var resultFromTypeScript = await Electron.HostHook.CallAsync&lt;string&gt;("create-excel-file", folderPath);
     Electron.IpcMain.Send(mainWindow, "excel-file-created", resultFromTypeScript);
 });</code>
                     </pre>


src/ElectronNET.ConsoleApp/ElectronNET.ConsoleApp.csproj
```diff
@@ -69,7 +69,7 @@
     <ProjectReference Include="..\ElectronNET.API\ElectronNET.API.csproj" Condition="$(ElectronNetDevMode)" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="ElectronNET.Core" Version="0.1.0" Condition="'$(ElectronNetDevMode)' != 'True'" />
+    <PackageReference Include="ElectronNET.Core" Version="0.0.18" Condition="'$(ElectronNetDevMode)' != 'True'" />
```
FlorianRappl, line 72: Wrong - always set to what `main` / the next release will be.

Also see condition. It should not be included in dev, if it is then your setup is wrong.

src/ElectronNET.WebApp/Views/About/Index.cshtml
```diff
@@ -20,7 +20,7 @@
 {
     return WebHost.CreateDefaultBuilder(args)
         .UseElectron(args)
-        .UseStartup<Startup>()
+        .UseStartup&lt;Startup>()
```
FlorianRappl: Would prefer the `>` to be also transformed (`&gt;`) - for consistency / clarity. Same for the other occurrence.

src/ElectronNET.ConsoleApp/ElectronNET.ConsoleApp.csproj
agracio: Not sure I am following you. Currently if I checkout `develop` branch .NET solution does not build and it is not possible to start demo project because version 0.1.0 does not exist in nuget. Latest version is `0.1.0-pre.139` is that what you mean?

FlorianRappl: You should either checkout `main` - as its the release branch - or set the `ElectronNetDevMode` property to `true`.

The default branch of the project is `main`. The NuGet references are supposed to work against the latest released one - so a snapshot of `main` for `0.1.0` should work against `0.1.0`. Therefore, `develop` has to work against the version that *will* be released. Hence NuGet references are not supposed to work in `develop` (as its pointing towards the future) - which is why `ElectronNetDevMode` is there.

Alternative variant is that it's always getting the packages from the local repository, however, I like that we can be flexible here with the `ElectronDevMode`.

We could switch its default - so instead of defaulting to `False` it could be `True`, i.e., it would always use the local ones (which are always right - as its in the same repository). In that scenario we would only switch it to `False` it we would like to test / verify a specific release or an issue with a release.

agracio: Still not entirely clear on what you mean.

As we discussed before `main` branch is broken and is awaiting fixes, it cannot be worked on. The last release against it was 0.0.18 and if I look at project references in code in `main`:

```xml
<PackageReference Include="ElectronNET.Core" Version="0.0.18" Condition="'$(ElectronNetDevMode)' != 'True'" />
<PackageReference Include="ElectronNET.Core.AspNet" Version="0.0.18" Condition="'$(ElectronNetDevMode)' != 'True'" />
```

Not sure what you mean by "always set to what `main` / the next release will be" - nuget reference cannot be set to an unreleased (future) version unless we assume that `develop` branch is not suppose to be able to run demo projects at all and solution compile is meant to be broken.

If we are setting to the last released version from `main` then it is 0.0.18.

Not sure how we can work on `develop` branch if it does not compile. If the changes to nuget packages are meant to be local only then it would make it hard to commit any other changes that might have been part of .csproj.

agracio: I can see now that switching `ElectronNetDevMode` allows for local references to work but having to switch it manually in .csproj file is not very intuitive to say the least.

agracio: Having said that so far was not able to run it locally using ElectronNetDevMode  set to True.

FlorianRappl: Might not be intuitive but I understood it with just one look. You might need two - but you still get the idea.

agracio: I was not able to run demo project after switching to 'True' yesterday but your new changes fixed everything. 
Really appreciate the work that you have done.

FlorianRappl: Well all of the work is from @softworkz - I am just putting the pieces together.

Anyway, thanks for your contribution!

softworkz: I think I need to change some notification settings - I haven't got any notification from Electron.NET rrecently.

Most ironically, Florian, it was your mention that seems to have triggered an e-mail, which had never got working on GitHub - that's sick...

PS: Haven't looked at the conversation and the PR yet 😆

FlorianRappl: I also send you an email to make dinner plans 🤗  @softworkz

softworkz: > I also send you an email to make dinner plans 🤗 @softworkz

Oh - now I see it - MS thought it's junk e-mail. Will respond shortly.

softworkz:
Hi,

let me explain those oddities you are seeing:

The primary goal for the PR was to provide a successful initial experience when testing it. By that I mean that it is working right out of the box: You open it in Visual Studio, you can compile it and run the samples without any hoops to jump through.
That's why I did something that you wouldn't do normally: I'v checked in ready-built nuget packages and made nuget.config to load the packages from the local folder. Just to save everybody from doing all the steps that are normally needed when you have a solution with projects that are self-consuming nuget packages produced by that same solution.
It was simply a shortcut for you to get going without further ado (like "it's not working" >> "yea, you gotta do 1, 2 and 3 first and then x and y...").

Enabling "dev mode" is naturally the way you want for developing. I remember a note from @GregorBiswanger, complaining about "excessive nuget caching", which is in fact a problem when you want to work in a solution which is both - producer+consumer of the same nuget package (background is that nuget packages are meant to be immutable, that's why Visual Studio doesn't monitor nuget packages for changes - like it does for almost all other - even external - sources).
So - the "dev mode" resolves that: All changes that you make are effective immediately.

But there's one caveat with that "dev mode": When it is enabled, the MSBuild props and targets are used from the in-project location - which also applies to the ElectronNET.Build.dll. This dll gets loaded by VS because the test projects consume the targets which use the custom tasks in this dll. In turn, the dll gets locked and when you try to build the solution, you get an error about not being able to write to the dll's location because it is locked (by VS).

What I do locally here, is to simply uncheck the build dll project in the solution configuration. But that's a change that I could not include in the PR, because when it's unchecked in the configuration, it also wouldn't get built when needed for building the nuget packages.

As you see, even though it's not overly complicated, it's a lot to explain...and I jsut wanted to skip that explanation and keep it for later (=now).

Thanks,
sw

Issue #906, opened 2025-11-06
Creating a window with contextIsolation enabled does not enable contextIsolation in electron


NimbusFox:
* **Version**: 0.0.18

.net 9

* **Target**: Linux and Windows (Issue was found on Linux)

In WindowManager.cs lines 128-144. If the X and Y values are -1 it uses the global json serializer else it will create a new serializer.

When it uses the global json serializer it fails to enable the contextIsolation and possibly other items as well. Where as if the X or Y value is not -1 it uses a local json serializer which passes along all the values as their defaults if they have not been set a value yet which triggers the options to be set correctly

FlorianRappl:
Yeah sounds reasonable. Mind doing a PR?

PR #907, opened 2025-11-06
Removed ownjsonSerializer from WindowManager

NimbusFox:
Removed the unneeded serializer and applied the settings to the global this._jsonSerializer as when the X and Y values are -1 in the options it would lead to some settings not being applied. E.g. ContextIsolation with it now applying the default values when serializing, settings mentioned before are correctly assigned on Electron's end

In regards to #906

### DIFF
--- a/src/ElectronNET.API/API/WindowManager.cs
+++ b/src/ElectronNET.API/API/WindowManager.cs
@@ -153,12 +153,7 @@ namespace ElectronNET.API
                     options.X -= 7;
                 }
 
-                var ownjsonSerializer = new JsonSerializer()
-                {
-                    ContractResolver = new CamelCasePropertyNamesContractResolver(),
-                    NullValueHandling = NullValueHandling.Ignore
-                };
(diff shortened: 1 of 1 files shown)


FlorianRappl (approved):
LGTM!

Any idea why we might have needed / used the previous behavior on the `_jsonSerializer` @GregorBiswanger ?

FlorianRappl:
(Wrong base branch used - I changed the target to `develop`, but that has no impact on the actions, i.e., it does not re-trigger the actions to run)

softworkz:
> Any idea why we might have needed / used the previous behavior on the `_jsonSerializer` @GregorBiswanger ?

The earlier SocketIO libraries didn't have built-in serialization.

FlorianRappl:
> The earlier SocketIO libraries didn't have built-in serialization.

Yes, but why was the behavior set to `DefaultValueHandling.Ignore`?

softworkz:
> > The earlier SocketIO libraries didn't have built-in serialization.
> 
> Yes, but why was the behavior set to `DefaultValueHandling.Ignore`?

Typically it would be about reducing payloads, but 

- that's not really a concern on local loopback networking
- it's not valid to do so across language boundaries, because the JS side will unlikely have the same default values (usually none at all)

PR #911, opened 2025-11-08
feat: updated 'TitleBarOverlay' property to be passed as object

Denny09310:
I’ve updated the configuration to allow titleBarOptions to be passed either as an object or a boolean, in line with the Electron documentation. I did some quick testing and it appears to be working correctly.

### DIFF
--- a/src/ElectronNET.API/Converter/TitleBarOverlayConverter.cs
+++ b/src/ElectronNET.API/Converter/TitleBarOverlayConverter.cs
@@ -26,6 +26,7 @@ public class TitleBarOverlayConverter : JsonConverter<TitleBarOverlay>
     {
         if (value is null)
         {
+            writer.WriteUndefined();
             return;
         }
 


src/ElectronNET.API/Converter/TitleBarOverlayConverter.cs
```diff
@@ -0,0 +1,42 @@
...
+        {
+            throw new JsonSerializationException("Invalid value for TitleBarOverlay. Expected true, false, or an object.");
+        }
+    }
+
+    public override void WriteJson(JsonWriter writer, TitleBarOverlay value, JsonSerializer serializer)
+    {
+        if (value is null)
+        {
+            writer.WriteNull();
```
FlorianRappl: Are you sure the converter is called in case of  `null`?

In case it's called I'd actually write `undefined` (`WriteUndefined()`). Reason: `null` may have a meaning (or may not be allowed) on the other side, while `undefined` will refer to the default value (i.e., it won't be transmitted via JSON serialization, so it will be treated like it's not there).

Denny09310: Yes, you're right. It was a C# 1:1 translation error 😅

FlorianRappl (commented):
Great addition - much appreciated!

src/ElectronNET.API/Converter/TitleBarOverlayConverter.cs
Denny09310: I don't think there is a WriteUndefined in System.Text.Json, I think not writing anything is like "undefined"

FlorianRappl: OK, but right now we are still using Newtonsoft.Json - which has this method. Once we transform (see discussion) this (among other behavior changes) will be taken into consideration.

Denny09310: Right 😅. I was fiddling with the System.Text.Json migration and got confused

PR #912, opened 2025-11-08
Demo fixes

Denny09310:
I'm not very familiar with the codebase yet, especially since I still need to go through all the TS/JS parts. While looking at the develop branch (the one without the migration to System.Text.Json), I added some comments to point out what isn't working in the demo. I figured that if someone with more experience in the codebase is interested, they could help fix those issues, so we can have a solid base to work from when attempting the migration in the future.

### DIFF
--- a/src/ElectronNET.API/Converter/TitleBarOverlayConverter.cs
+++ b/src/ElectronNET.API/Converter/TitleBarOverlayConverter.cs
@@ -26,6 +26,7 @@ public class TitleBarOverlayConverter : JsonConverter<TitleBarOverlay>
     {
         if (value is null)
         {
+            writer.WriteUndefined();
             return;
         }
 


Denny09310:
This where quick findings, I'll try debug them later

src/ElectronNET.API/Runtime/Services/LifetimeServiceBase.cs
```diff
@@ -81,6 +81,7 @@ protected void TransitionState(LifetimeState newState)
                 return;
             }
 
+            // FIXME: At launch the newState is 'Started' and this.state is 'Ready' throwing always an error
```
softworkz: Should be fixed here: https://github.com/ElectronNET/Electron.NET/pull/913/commits/9c1b688dd16afb7146d2364a668e603ef0e5c5f4

src/ElectronNET.WebApp/Controllers/AppSysInformationController.cs
```diff
@@ -11,6 +11,7 @@ public IActionResult Index()
         {
             if(HybridSupport.IsElectronActive)
             {
+                // FIXME: System.TimeoutException: 'No response after 1000ms trying to retrieve value app.GetAppPathAsync()'
```
softworkz: Should be fixed by this: https://github.com/ElectronNET/Electron.NET/pull/913

src/ElectronNET.WebApp/Controllers/UpdateController.cs
```diff
@@ -27,6 +27,8 @@ public IActionResult Index()
                 };
                 Electron.AutoUpdater.OnUpdateDownloaded += async (info) => await Electron.Dialog.ShowMessageBoxAsync("Update complete!" + info.Version);
 
+
+                // FIXME: System.TimeoutException: 'No response after 1000ms trying to retrieve value app.GetVersionAsync()'
```
softworkz: Should be fixed here: https://github.com/ElectronNET/Electron.NET/pull/913

PR #916, opened 2025-11-09
Fix link to 'What's New in ElectronNET.Core'

damianh:
The link in the readme is resolving to `https://github.com/ElectronNET/Electron.NET/blob/main/wiki/What's-New`, it should be resolving to `https://github.com/ElectronNET/Electron.NET/wiki/What's-New`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 
 ElectronNET.Core represents a fundamental modernization of Electron.NET, addressing years of accumulated pain points while preserving full API compatibility. This isn't just an update—it's a complete rethinking of how .NET developers build and debug cross-platform desktop applications with Electron.
 
-Read more: [**What's New in ElectronNET.Core**](wiki/What's-New)
+Read more: [**What's New in ElectronNET.Core**](../../wiki/What's-New)
 
 
 Build cross platform desktop applications with .NET 6/8 - from console apps to ASP.Net Core (Razor Pages, MVC) to Blazor


README.md
```diff
@@ -8,7 +8,7 @@
 
 ElectronNET.Core represents a fundamental modernization of Electron.NET, addressing years of accumulated pain points while preserving full API compatibility. This isn't just an update—it's a complete rethinking of how .NET developers build and debug cross-platform desktop applications with Electron.
 
-Read more: [**What's New in ElectronNET.Core**](wiki/What's-New)
+Read more: [**What's New in ElectronNET.Core**](../../wiki/What's-New)
```
softworkz, line 11:
```suggestion
Read more: [**What's New in `ElectronNET.Core`**](https://github.com/ElectronNET/Electron.NET/wiki/What's-New)
```

It needs an absolute URL, like I've done in the ReadMe.md in the develop branch. 

The reason is that the ReadMe can be shown from different URLs, that's why a relative path will always be broken in one or another case.

FlorianRappl: Yes, that's right. One example is the README presentation on [NuGet](https://www.nuget.org/packages/ElectronNET.Core/0.1.0-pre.171#readme-body-tab).

FlorianRappl:
PR had wrong target (and most likely a wrong base branch - see target branch).

FlorianRappl:
Actually, as it's the only change in the PR I'm going to close it. This was fixed already in the `develop` branch.

PR #917, opened 2025-11-09
refactor: Migrating from Newtonsoft.Json to System.Text.Json

Denny09310:
This PR replaces Newtonsoft.Json with the built-in System.Text.Json. The tests were taken from @softworkz’s branch and all are passing, except for the two already noted in issue #913.

### DIFF
--- a/src/ElectronNET.API/API/App.cs
+++ b/src/ElectronNET.API/API/App.cs
@@ -547,11 +547,11 @@ namespace ElectronNET.API
             var taskCompletionSource = new TaskCompletionSource<string>();
             using (cancellationToken.Register(() => taskCompletionSource.TrySetCanceled()))
             {
-                BridgeConnector.Socket.On<JsonElement>("appGetPathCompleted", (path) =>
+                BridgeConnector.Socket.On<string>("appGetPathCompleted", (path) =>
                 {
                     BridgeConnector.Socket.Off("appGetPathCompleted");
 
-                    taskCompletionSource.SetResult(path.GetString());
+                    taskCompletionSource.SetResult(path);
                 });
 
                 BridgeConnector.Socket.Emit("appGetPath", pathName.GetDescription());
@@ -716,10 +716,10 @@ namespace ElectronNET.API
             var taskCompletionSource = new TaskCompletionSource<bool>();
             using (cancellationToken.Register(() => taskCompletionSource.TrySetCanceled()))
             {
-                BridgeConnector.Socket.On<JsonElement>("appSetAsDefaultProtocolClientCompleted", (success) =>
+                BridgeConnector.Socket.On<bool>("appSetAsDefaultProtocolClientCompleted", (success) =>
                 {
                     BridgeConnector.Socket.Off("appSetAsDefaultProtocolClientCompleted");
-                    taskCompletionSource.SetResult(success.GetBoolean());
+                    taskCompletionSource.SetResult(success);
                 });
 
                 BridgeConnector.Socket.Emit("appSetAsDefaultProtocolClient", protocol, path, args);
@@ -770,10 +770,10 @@ namespace ElectronNET.API
             var taskCompletionSource = new TaskCompletionSource<bool>();
             using (cancellationToken.Register(() => taskCompletionSource.TrySetCanceled()))
             {
-                BridgeConnector.Socket.On<JsonElement>("appRemoveAsDefaultProtocolClientCompleted", (success) =>
+                BridgeConnector.Socket.On<bool>("appRemoveAsDefaultProtocolClientCompleted", (success) =>
                 {
                     BridgeConnector.Socket.Off("appRemoveAsDefaultProtocolClientCompleted");
-                    taskCompletionSource.SetResult(success.GetBoolean());
+                    taskCompletionSource.SetResult(success);
                 });
 
                 BridgeConnector.Socket.Emit("appRemoveAsDefaultProtocolClient", protocol, path, args);
@@ -842,10 +842,10 @@ namespace ElectronNET.API
             var taskCompletionSource = new TaskCompletionSource<bool>();
             using (cancellationToken.Register(() => taskCompletionSource.TrySetCanceled()))
             {
-                BridgeConnector.Socket.On<JsonElement>("appIsDefaultProtocolClientCompleted", (success) =>
+                BridgeConnector.Socket.On<bool>("appIsDefaultProtocolClientCompleted", (success) =>
                 {
                     BridgeConnector.Socket.Off("appIsDefaultProtocolClientCompleted");
-                    taskCompletionSource.SetResult(success.GetBoolean());
+                    taskCompletionSource.SetResult(success);
                 });
 
                 BridgeConnector.Socket.Emit("appIsDefaultProtocolClient", protocol, path, args);
@@ -870,10 +870,10 @@ namespace ElectronNET.API
             var taskCompletionSource = new TaskCompletionSource<bool>();
             using (cancellationToken.Register(() => taskCompletionSource.TrySetCanceled()))
             {
-                BridgeConnector.Socket.On<JsonElement>("appSetUserTasksCompleted", (success) =>
+                BridgeConnector.Socket.On<bool>("appSetUserTasksCompleted", (success) =>
                 {
                     BridgeConnector.Socket.Off("appSetUserTasksCompleted");
-                    taskCompletionSource.SetResult(success.GetBoolean());
+                    taskCompletionSource.SetResult(success);
                 });
 
                 BridgeConnector.Socket.Emit("appSetUserTasks", userTasks);
@@ -943,10 +943,10 @@ namespace ElectronNET.API
             var taskCompletionSource = new TaskCompletionSource<bool>();
             using (cancellationToken.Register(() => taskCompletionSource.TrySetCanceled()))
             {
-                BridgeConnector.Socket.On<JsonElement>("appRequestSingleInstanceLockCompleted", (success) =>
+                BridgeConnector.Socket.On<bool>("appRequestSingleInstanceLockCompleted", (success) =>
                 {
                     BridgeConnector.Socket.Off("appRequestSingleInstanceLockCompleted");
-                    taskCompletionSource.SetResult(success.GetBoolean());
+                    taskCompletionSource.SetResult(success);
                 });
 
                 BridgeConnector.Socket.Off("secondInstance");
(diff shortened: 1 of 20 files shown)


FlorianRappl:
Can you resolve the conflicts? I know its annoying, but right now due to the conflicts it does not show the true diff (many of the changes are already in `develop` as you took the other branch before it landed in `develop`). Many thanks!

Denny09310:
Yes, absolutely!

softworkz:
@Denny09310  - this doesn't look right. The PR is still including my commits. You need to **rebase** onto the develop branch, instead of submitting with merge commit.

Denny09310:
Hope to have done everything right. Sorry but I don't use GitHub at this level frequently so I don't know how to do everything 😅

softworkz:
> Hope to have done everything right. Sorry but I don't use GitHub at this level frequently so I don't know how to do everything 

All perfect now, thanks!

src/ElectronNET.API/ElectronNET.API.csproj
```diff
@@ -27,8 +27,6 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
     </PackageReference>
-    <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
-    <PackageReference Include="SocketIO.Serializer.NewtonsoftJson" Version="3.1.2" />
```
FlorianRappl, line 31: I thought we might want to use ` SocketIO.Serializer.SystemTextJson`? I see a lot of code skipping automatic serialization / deserizaliation so that might be worth it.

softworkz:
What was it, what made the one test fail? 
Not that I'm really interested in that specific detail, it's just about whether it was an issue that could possibly also bite in other cases?

src/ElectronNET.API/Converter/PageSizeConverter.cs
```diff
@@ -1,43 +1,47 @@
...
+        else if (reader.TokenType == JsonTokenType.StartObject)
         {
-            return serializer.Deserialize<PageSize>(reader);
+            using var doc = JsonDocument.ParseValue(ref reader);
+            return doc.RootElement.Deserialize<PageSize>(API.Serialization.ElectronJson.Options);
         }
         else
         {
-            throw new JsonSerializationException("Invalid value for PageSize. Expected true, false, or an object.");
+            throw new JsonException("Invalid value for PageSize. Expected string or an object.");
```
FlorianRappl, line 23: Great catch there!

src/ElectronNET.API/ElectronNET.API.csproj
Denny09310: It's not already included? I should have already set the serializer as `SystemTextJsonSerializer`

>public SocketIoFacade(string uri)
>{
>     _socket = new SocketIO(uri);
>    _socket.Serializer = new SystemTextJsonSerializer(ElectronJson.Options);
>     // Use default System.Text.Json serializer from SocketIOClient.
>     // Outgoing args are normalized to camelCase via SerializeArg in Emit.
>}

src/ElectronNET.API/Converter/PageSizeConverter.cs
```diff
@@ -1,43 +1,47 @@
...
         }
     }
 
-    public override void WriteJson(JsonWriter writer, PageSize value, JsonSerializer serializer)
+    public override void Write(Utf8JsonWriter writer, PageSize value, JsonSerializerOptions options)
     {
         if (value is null)
         {
-            writer.WriteUndefined();
+            writer.WriteNullValue();
```
FlorianRappl: I would much rather prefer not to write anything. If we can't specify undefined we must find another way.

`null` = transported via JSON and might have a different meaning
`undefined` = not transported via JSON - property is omitted and therefore default is taken for sure

src/ElectronNET.API/Converter/TitleBarOverlayConverter.cs
```diff
@@ -1,43 +1,46 @@
...
         }
     }
 
-    public override void WriteJson(JsonWriter writer, TitleBarOverlay value, JsonSerializer serializer)
+    public override void Write(Utf8JsonWriter writer, TitleBarOverlay value, JsonSerializerOptions options)
     {
         if (value is null)
         {
-            writer.WriteUndefined();
+            writer.WriteNullValue();
```
FlorianRappl: See other comment - we need a way to preserve this meaning. `undefined` was written for a reason. Writing `null` will change semantics.

Denny09310:
> What was it, what made the one test fail? Not that I'm really interested in that specific detail, it's just about whether it was an issue that could possibly also bite in other cases?

It was `Notification_create_check` not calling the `OnShow` due to `NotificationOptions.ShowID` not being serialized. Fixed by adding [JsonInclude] attribute

src/ElectronNET.API/API/Entities/ThumbarButton.cs
```diff
@@ -36,7 +35,7 @@ public class ThumbarButton
         /// hidden - The button is not shown to the user.
         /// noninteractive - The button is enabled but not interactive; no pressed button state is drawn.This value is intended for instances where the button is used in a notification.
         /// </summary>
-        [JsonProperty("flags", ItemConverterType = typeof(StringEnumConverter))]
+        [JsonPropertyName("flags")]
         public ThumbarButtonFlag[] Flags { get; set; }
```
softworkz: Why is this attribute needed? Isn't camelCase set globally?

src/ElectronNET.API/API/Entities/RemovePassword.cs
```diff
@@ -28,7 +28,7 @@ public class RemovePassword
         /// Scheme of the authentication. Can be basic, digest, ntlm, negotiate. 
         /// Must be provided if removing by origin.
         /// </summary>
-        [JsonConverter(typeof(StringEnumConverter))]
+        [JsonConverter(typeof(JsonStringEnumConverter))]
```
softworkz: Can't this be covered by a global serializer setting?
I've never seen this anywhere (unless the global setting is by number and you have a few exceptions..

src/ElectronNET.API/ElectronNET.API.csproj
softworkz: SocketIO.Serializer.SystemTextJson is the default now for SocketIO. It's a dependency anyway, so it needs no PackageReference.

src/ElectronNET.API/API/App.cs
```diff
@@ -1071,13 +1068,13 @@ public async Task<int> ImportCertificateAsync(ImportCertificateOptions options,
             var taskCompletionSource = new TaskCompletionSource<int>();
             using (cancellationToken.Register(() => taskCompletionSource.TrySetCanceled()))
             {
-                BridgeConnector.Socket.On("appImportCertificateCompleted", (result) =>
+                BridgeConnector.Socket.On<JsonElement>("appImportCertificateCompleted", (result) =>
                 {
                     BridgeConnector.Socket.Off("appImportCertificateCompleted");
-                    taskCompletionSource.SetResult((int)result);
+                    taskCompletionSource.SetResult(result.GetInt32());
                 });
```
softworkz: Why do you use JsonElement everywhere.  Doesn't this work?

```csharp
BridgeConnector.Socket.On<int>("appImportCertificateCompleted", (result) =>
```

If not, why?

src/ElectronNET.API/ElectronNET.API.csproj
FlorianRappl: OK then I fully agree with your question that `JsonElement` should not be used as the target type. It would do a proper deserialization anyway.

src/ElectronNET.API/API/Entities/RemovePassword.cs
Denny09310: Yes, it's already done by `ElectronJson.Options`, I think I can just remove them:

```cs
public static readonly JsonSerializerOptions Options = new JsonSerializerOptions
{
    PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
    DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull,
    WriteIndented = false,
    Converters =
    {
        new JsonStringEnumConverter(JsonNamingPolicy.CamelCase)
    }
};
```

Denny09310:
@softworkz @FlorianRappl Okay, probably something I was doing wrong long before this PR, I'm going to replace all the `.On<JsonElement>` into the corresponding value and test it out

Denny09310:
As the commit description says, I've refactored all `On` calls to deserialize directly to the correct type except some methods that expects mixed arrays or complex values (Eg. AddEvent args). Any ideas or we leave them as they are? @softworkz @FlorianRappl

FlorianRappl:
> Any ideas or we leave them as they are?

Of course if we know the `T` we could pass it on - otherwise, especially for the complex / mixed case, I'd say it's good to leave them at the JSON representation (`JsonElement`).

FlorianRappl (approved):
LGTM - thanks for the superb work @Denny09310 !

agracio:
@Denny09310, @FlorianRappl  There is an error in this PR.

IpcMain.cs
```cs
public void Send(BrowserWindow browserWindow, string channel, params object[] data)
{
    BridgeConnector.Socket.Emit("sendToIpcRenderer", browserWindow, channel, data);
}
```

This causes data to always be sent as array, to reproduce fire a demo app -> "Get app and system information" -> "Get screen information". 

This will fail to show correct display size, a quick debug shows that IpcMain sends `[Size]` instead of `Size`.

Demo app code:
```cs
Electron.IpcMain.On("screen-info", async (args) =>
{
    var display = await Electron.Screen.GetPrimaryDisplayAsync();
    var mainWindow = Electron.WindowManager.BrowserWindows.First();
    Electron.IpcMain.Send(mainWindow, "got-screen-info", display.Size);
});
```

FlorianRappl:
Great catch. Yes, needs to be fixed. Thanks for spotting.

agracio:
> Great catch. Yes, needs to be fixed. Thanks for spotting.

False alarm, this is not a new bug and was present before the change to `Send` method. Not sure if it was always broken or is a part of another PR. Will continue checking various builds.

agracio:
I ran all version all the way back to start of refactoring, all have the same problem with sending `Size` object as an array.
Pretty sure that it should be working.
It looks like the issue does not affect primitive types like strings or ints.

Tried everything but cannot figure out the reason for this behaviour.

softworkz:
@agracio - Thanks for the invesigation. Fixed here: https://github.com/ElectronNET/Electron.NET/pull/922

PR #930, opened 2025-11-15
build project: Update packages (due to vulnerabilities)

softworkz:
This does just a minimal bump for the transitive packages, at the same api level, just a non-vulnerable revision


FlorianRappl (approved):
LGTM!

PR #937, opened 2025-11-17
Move PR comment creation to separate workflow on:pull_request_target

softworkz:


softworkz:
Should be ok and safe now.

Issue #900, opened 2025-11-01
Broken build after migration: error MSB4184: The expression "[System.IO.Path]::GetFullPath('')"


mu88:
* **Version**: `v0.0.18`

* **Target**: `win-x64`

Steps to Reproduce:

1. Clone the repository https://github.com/mu88/BlazorFotoManager
2. Switch to the feature branch `feature/new-api`
3. Build the app via `dotnet build`
4. Publish the app via `dotnet publish ./FotoManager/FotoManager.csproj -p:PublishProfile=win-x64`

The build fails with `D:\Development\PackageCaches\NuGet\electronnet.core\0.0.18\build\ElectronNET.LateImport.targets(431,7): error MSB4184: The expression "[System.IO.Path]::GetFullPath('')" cannot be evaluated. The path is empty. (Parameter 'path')`

<img width="2560" height="360" alt="Image" src="https://github.com/user-attachments/assets/674e664e-bc4d-46cb-a49c-b1d9cd3d7f9b" />

When using the old API via `ElectronNET.API`, I can successfully build my app.

FlorianRappl:
Fixed in 0.1.0.

FlorianRappl:
@mu88 while this was monkey patched in 0.1.0 and 0.2.0 for 0.3.0 the error will come back.

See https://github.com/ElectronNET/Electron.NET/pull/942#issuecomment-3552022687 for details on how to resolve that on your side.

PR #942, opened 2025-11-18
electron-builder: Provide correct output path

softworkz:
This had gotten lost somehow in my original PR.

### DIFF
--- a/src/ElectronNET/build/ElectronNET.LateImport.targets
+++ b/src/ElectronNET/build/ElectronNET.LateImport.targets
@@ -344,7 +344,6 @@
     <PropertyGroup>
       <_OriginalPublishDir>$(PublishDir)</_OriginalPublishDir>
       <PublishDir>$(_OriginalPublishDir)bin\</PublishDir>
-      <PublishUrl>$(PublishDir)</PublishUrl>
       <DeleteExistingFiles>true</DeleteExistingFiles>
     </PropertyGroup>
 
@@ -460,7 +459,7 @@
     </PropertyGroup>
 
     <PropertyGroup>
-      <_NpxCmd>npx electron-builder --config=./$(ElectronBuilderJson) --$(ElectronPlatform) --$(ElectronArch) -c.electronVersion=$(ElectronVersion) $(ElectronPaParams)</_NpxCmd>
+      <_NpxCmd>npx electron-builder --config=./$(ElectronBuilderJson) --$(ElectronPlatform) --$(ElectronArch) -c.electronVersion=$(ElectronVersion) -c.directories.output &quot;$(ElectronPublishUrlFullPath)&quot; $(ElectronPaParams)</_NpxCmd>
       <_NpxCmd Condition="'$(IsLinuxWsl)' == 'true'">wsl bash -ic '$(_NpxCmd)'</_NpxCmd>
     </PropertyGroup>
 


softworkz:
@FlorianRappl - Can we merge this? It broke my build (after switching to our new packages from here)...

src/ElectronNET/build/ElectronNET.LateImport.targets
```diff
@@ -460,7 +460,7 @@
     </PropertyGroup>
 
     <PropertyGroup>
-      <_NpxCmd>npx electron-builder --config=./$(ElectronBuilderJson) --$(ElectronPlatform) --$(ElectronArch) -c.electronVersion=$(ElectronVersion) $(ElectronPaParams)</_NpxCmd>
+      <_NpxCmd>npx electron-builder --config=./$(ElectronBuilderJson) --$(ElectronPlatform) --$(ElectronArch) -c.electronVersion=$(ElectronVersion) -c.directories.output &quot;$(ElectronPublishUrlFullPath)&quot; $(ElectronPaParams)</_NpxCmd>
```
FlorianRappl, line 462: I think I removed that (need to verify). For me this resulted in a recursive loop as the output was in the same directory as the signing. That was really bad.

Since the output is anyway placed in the CWD (which is the publish directory) I don't see the direct reason for this. For me, the publish *only* worked when this was removed. Maybe it was an edge case, but I used a new project with a default publish config.

softworkz: The point is that the output should go in the directory that you specify in the publish profile.

image

When I publish the WebApp with the included profiles, all is working fine. In which configuration did it fail for you?
(electron-builder)?

When the output is not specified, it writes the output into a "dist" folder instead of the publish output folder.

softworkz: Maybe you ran publish without profile and without specifying a publish output folder?

FlorianRappl: No I published with a profile. And it published to the given directory, however, it got stuck in the process (as described) and the signing tool started go berserk (only seen with verbose flags provided, otherwise all of that is hidden).

The error is really nasty and as written you don't need that flag. The PublishUrl we use to specify the CWD, so I don't see why we should double it. For me it only caused havoc.

FlorianRappl: Even if I would not have published without a profile I don't why we should specify this. The error is really nasty as it does not only stay hidden (and fails to complete the publish process), it also eats up essentially all the disk space if not stopped in time.

Again, as the publish is done in the CWD and the electron-builder just works as designed I don't see why we should be explicit here. Have you maybe skipped signing?

softworkz: Ah, I think it's when 

    `<ElectronNetDevMode>true</ElectronNetDevMode>`

is enabled.

FlorianRappl:
> an we merge this? It broke my build (after switching to our new packages from here)...

No, this breaks my publish process (recursive creation of directories & files). The error here is really nasty. We first need to check if this is still the case.

Can you specify how it breaks your build? The electron-builder runs in the CWD specified as the publish directory and for me this is all that's needed. It will work with the defaults just fine. What exactly is breaking if used with the defaults?

softworkz:
> Can you specify how it breaks your build? The electron-builder runs in the CWD specified as the publish directory and for me this is all that's needed. It will work with the defaults just fine. What exactly is breaking if used with the defaults?

That's a bit tricky. There is a fundemental different between MBuild publish for AspNetCore and a normal projects. That's why we're seeing different things.

softworkz:
It's also not the same path. One is ElectronPublishDirFullPath and the other is ElectronPublishUrlFullPath which are IIRC same in one case and different in the other.,

> No, this breaks my publish process (recursive creation of directories & files). T

Using the actual packages or with ElectronNetDevMode set true?

softworkz:
With the current package, when publishing from VS (console app), it goes into:

`project\obj\x64\Release\net8.0\linux-x64\PubTmp\dist\`

instead of here:

`project\publish\Release\net8.0\linux-x64\`

FlorianRappl:
> Using the actual packages or with ElectronNetDevMode set true?

Actual packages.

> There is a fundemental different between MBuild publish for AspNetCore and a normal projects. That's why we're seeing different things.

That's true. My scenario was an ASP.NET Core project. Presumably, we maybe should add a flag / only use it if the directories won't match / overlap?

src/ElectronNET/build/ElectronNET.LateImport.targets
FlorianRappl: I have seen this with the actual packages, but I guess / fear that with the dev mode it might also be true. TBH I think the actual problem is within the electron-builder. Even setting some directories as ignored did not help. It should never act like this. But we can't patch / change the electron-builder.

softworkz:
Let me check what's happening. It has been working fine for a long time. I've switched to the console model just very recently.

I see what's happening. Nested folder ad infinity - explorer and VS are crashing...

FlorianRappl:
> I see what's happening. Nested folder ad infinity - explorer and VS are crashing...

Yes, that is what occurred on my system. Did I mention that this was "nasty"? Windows actually had problems deleting the folder as the depth was too high...

softworkz:
> > I see what's happening. Nested folder ad infinity - explorer and VS are crashing...
> 
> Yes, that is what occurred on my system. Did I mention that this was "nasty"? Windows actually had problems deleting the folder as the depth was too high...

Right. Even `rd /S` has crashed cmd.exe 😆 

I don't know whether I've ever seen cmd itself crashing...

But PowerShell did it.

One thought: Maybe it's Windows-specific. I've tested all the publishing before submitting the first PR, but maybe not Windows as target.  

What is this signing anyway? What does it want to sign and how - without a cert?

FlorianRappl:
I am 100% sure its Windows specific. My guess is that electron-builder has a problem with the backslashes and compares unix formatted paths with the Windows paths. In Node.js both are equivalent, but if you string compare... well - then two equal paths are no longer equal.

softworkz:
You made another change:

image

That's what broke it and caused the recursion. The recursion came from the woing path and this parameter:  `-c.extraResources "bin/**/*"`  (which is right, though)

softworkz:
I've updated this PR accordingly.

src/ElectronNET/build/ElectronNET.LateImport.targets
```diff
@@ -344,7 +344,6 @@
     <PropertyGroup>
       <_OriginalPublishDir>$(PublishDir)</_OriginalPublishDir>
       <PublishDir>$(_OriginalPublishDir)bin\</PublishDir>
-      <PublishUrl>$(PublishDir)</PublishUrl>
```
softworkz, line 347: Adding this is what broke it and caused the recursion

FlorianRappl: Actually we need to ensure that `_OriginalPublishDir` ends with a (back)slash - otherwise the path (`PublishDir`) is really odd.

I'd say let's merge and test with this.

softworkz: > Actually we need to ensure that _OriginalPublishDir ends with a (back)slash - otherwise the path (PublishDir) is really odd.

Have you seen this happen? I think at the time when it is set from PublishDir, the PublishDir variable should already have been sanitized by MSBuild operations.

> I'd say let's merge and test with this.

Awesome, thank, I really need to get a new build out...

softworkz:
### Update

- Everything good on my side - the pipelines are creating the right output again now
- I tried publish of the WebApp after removing the trailing backslash in the `PublishUrl` => no problem
- Yet I noticed a new issue: I have broke something around HostHook with the recent project (and compilation) change for ElectronNET.Host
  need to investigate further...

FlorianRappl:
@softworkz 

Originally that change originated from #900 - I just tested the latest release (preview version) and the issue is back with your changes. Can you have a look?

FlorianRappl:
```
$> dotnet publish ./FotoManager/FotoManager.csproj -p:PublishProfile=win-x64
Restore complete (4,2s)
  FotoManagerLogic succeeded (0,2s) → FotoManagerLogic\bin\Release\net9.0\FotoManagerLogic.dll
  FotoManager failed with 1 error(s) (13,0s) → FotoManager\publish\Release\net9.0\win-x64\bin\
    C:\Users\Hans\.nuget\packages\electronnet.core\0.2.0-pre.262\build\ElectronNET.LateImport.targets(430,7): error MSB4184: The expression "[System.IO.Path]::GetFullPath('')" cannot be evaluated. The path is empty. (Parameter 'path')

Build failed with 1 error(s) in 26,1s

Workload updates are available. Run `dotnet workload list` for more information.
$> more .\FotoManager\FotoManager.csproj
<Project Sdk="Microsoft.NET.Sdk.Web">
  <PropertyGroup>
    <Authors>mu88</Authors>
    <Copyright>Copyright c $([System.DateTime]::Now.ToString('yyyy'))</Copyright>
    <RuntimeIdentifier>win-x64</RuntimeIdentifier>
    <NoWarn>CA1416</NoWarn>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="ElectronNET.Core" Version="0.2.0-pre.262"/>
    <PackageReference Include="ElectronNET.Core.AspNet" Version="0.2.0-pre.262"/>
    <PackageReference Include="System.Drawing.Common" Version="9.0.8"/>
  </ItemGroup>
  <ItemGroup>
    <ProjectReference Include="..\FotoManagerLogic\FotoManagerLogic.csproj"/>
  </ItemGroup>
  <ItemGroup>
    <Compile Remove="publish\**"/>
    <Content Remove="publish\**"/>
    <EmbeddedResource Remove="publish\**"/>
    <None Remove="publish\**"/>
    <_ContentIncludedByDefault Remove="publish\**"/>
    <None Update="assets\Icon.ico">
      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
    </None>
    <Folder Include="assets\"/>
    <Folder Include="wwwroot\"/>
  </ItemGroup>
</Project>
```

softworkz:
Yup, will check

softworkz:
Alright, in that project (https://github.com/mu88/BlazorFotoManager), the publish config in .pubxml is wrong. Should just copy ours.

Also, the electron-builder.json had a lot of things it should not have. I think we need to better document the migration. I thought it would be fine to say they need to copy the "build" subtree from package.json, but we also need to explain what needs to be removed.
(or maybe even add checks for things that should not be there)

Here's a fixed version of the foto manager which publishes fine with our latest packages: https://github.com/softworkz/BlazorFotoManager/tree/fix_electronnet

softworkz:
To be more specific: The publish profile in this project was wither copied from our console test app or from another non-WebSDK project. These are using `PublishDir` and `PublishProtocol`, but the publishing for WebSDK projects used `PublishURL` - that's why it was empty and it seemed you could fix it by setting `PublishURL` to `PublishDir`.

The logic in the LateImport.targets file is admittedly hard to follow. A prerequisite for that is to know which targets are run for one or the other publishing method and which run in both cases. And the order of execution. Then it's pretty easy to understand :-)

FlorianRappl:
Yeah I assumed already its outdated / wrong, but I think we should handle this case better.

For instance, if a specific variable is not set we should actually error out and print a good message. Otherwise, more people might fall into that trap (e.g., during their migration).

Example: Should we check if `PublishDir` is non-empty? If it is, we could error out / tell the user that the publish profile seems odd.

softworkz:
> Example: Should we check if `PublishDir`

Yes, absolutely but only in the case of WebSDK publishing. Somewhere there's a variable being set whether it's WebSDK or not. That would be the condition.
But the above is not a migration issue. 


Migration issues would be about electron-builder.json. There we need to check for all folder-related values to make sure none are specified.

softworkz:
> Migration issues would be about electron-builder.json. There we need to check for all folder-related values to make sure none are specified.

Especially everything with relative parent paths like `"..\asd..."`. It's very tempting to do so for getting something from the project, but this will blow up as soon as the built output is going to a different folder - like when publishing.

So the pattern is always like:

- When you want to reference something in the project (from electron-builder.json)
- Make sure it's copied to the build output
- Then reference it with a path relative to the build output

markatosi:
I attempted publishing on my mac and I saw similar behavior so this is not windows specific. (I'm using 0.2) It didn't create an infinite number of sub directories... it's stops eventually when the app bundle file has zero bytes. The top level folder where the first bundle lives is about 4 times as big as it should be. If you descend into the next directory the bundle is slightly smaller and so on and on and on. The interesting thing is the top level app bundle runs as you would expect. 

if i just build a release version and run electron-builder manually from the command line it works the way it's supposed to.

softworkz:
@markatosi - please check out my commit here: https://github.com/mu88/BlazorFotoManager/compare/main...softworkz:BlazorFotoManager:fix_electronnet to see how I fixed that other project.

markatosi:
Thank you. I'll check it out in a bit, time for lunch :-) I'll let you know if it resolves my issue soon.

No luck so far... I'm going to create a mostly empty blazor app and try and get this working  instead of using a copy of my real app to make sure there isn't something else going on.

markatosi:
The bare bones demo app published perfectly. 
I used the same publish profile xml and electron-builder plus my custom package.json.

On my production app i removed the bin and obj directories. That did nothing. 
Deleted the node_modules folder. Now publishing works. 

I need a beer. 

It's always the most obvious thing that i ignore

PR #945, opened 2025-11-20
BrowserWindowOptions; Fix default  and add more Tests

softworkz:
Electron's default is trueElectron's default for FullScreenable is true

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -22,7 +22,7 @@ jobs:
             rid: linux-x64
           - os: windows-2022
             rid: win-x64
-          - os: macos-14
+          - os: macos-26
             rid: osx-arm64
 
     env:


softworkz:
This  probably the explanation for why that serialization option "IgnoreDefaults" has been there.

I'm not sure whether this is already the right thing. Probably there are types of windows (like tool window or dialog) which may not be "FullScreenable" and Electron might error.
That would be my guess due to the fact that all other properties have defauilt values defined - but not this one.

I think the most safe and correct way would be to make all non-string properties nullable. This comes closest to the JS development behavior where you just omit those properties that you don't want to set.

Right now we need to see which defaults Electron has and set them accordingly. But when there's no constant default, this doesn't work out. Most likely, that's where somebody came up with the "IgnoreDefault" serialization option...

Issue #944, opened 2025-11-20
Maybe a bug. Electron.Screen.GetPrimaryDisplayAsync();


markatosi:
* **Version**: 0.2

* **Target**: .net 10.0, electron 38.2.2 node v22.12.0

Steps to Reproduce:

1. Try to use await Electron.Screen.GetPrimaryDisplayAsync();
2. Error:System.TimeoutException: No response after 1000ms trying to retrieve value screen.GetPrimaryDisplayAsync()

The Screen id is set to -1 so that's probably not correct.

softworkz:
I'm not sure how you got to those results.

The `GetPrimaryDisplayAsync()` method is tested by  the `Primary_display_has_positive_dimensions` test which is succeeding on all platforms (e.g. https://github.com/ElectronNET/Electron.NET/actions/runs/19494438431)

markatosi:
I have no idea either :-) All I did was migrate from electron-sharp to ElectronNet Core. 
I did test on MacOS 26.1 and 15.4.

I also tested the bare bones demo app. Doesn't work on that one either.

Is there anything you can recommend that I do to prove that it's not working for me?

softworkz:
Is it like that's the one thing not working and everything else is fine, or is this the first thing  your app does and you didn't look any further?

> Is there anything you can recommend that I do to prove that it's not working for me?

A minimal app which exposes that problem would be helpful to look at what you are doing. (unless your code is open anyway.. :-)

markatosi:
That was the only thing that I noticed wasn't working. Everything else has been fine. Code isn't open but the minimal app i can post

softworkz:
> Is there anything you can recommend that I do to prove that it's not working for me?

You could check out the develop branch and run dotnet test on that machine, so that we have a baseline (known to be working) for comparison.

softworkz:
That includes the test for the screen.API.

markatosi:
Full error from the minimal app. 

```
warn: Microsoft.AspNetCore.Components.Server.Circuits.RemoteRenderer[100]
      Unhandled exception rendering component: No response after 1000ms trying to retrieve value screen.GetPrimaryDisplayAsync()
      System.TimeoutException: No response after 1000ms trying to retrieve value screen.GetPrimaryDisplayAsync()
         at BlazorApp1.Components.Pages.Counter.IncrementCount() in /Volumes/Repos_APFS/source/electron/BlazorApp1/Components/Pages/Counter.razor:line 18
[...]

PR #949, opened 2025-11-21
Update all model classes to Electron API 39.2

softworkz:
-

#### ...and to electron-updater 6.7.2

### DIFF
--- a/src/ElectronNET.API/API/ApiBase.cs
+++ b/src/ElectronNET.API/API/ApiBase.cs
@@ -117,13 +117,18 @@ namespace ElectronNET.API
 
         protected Task<T> InvokeAsync<T>(object arg = null, [CallerMemberName] string callerName = null)
         {
+            return this.InvokeAsyncWithTimeout<T>(InvocationTimeout, arg, callerName);
+        }
+
+        protected Task<T> InvokeAsyncWithTimeout<T>(int invocationTimeout, object arg = null, [CallerMemberName] string callerName = null)
+        {
             Debug.Assert(callerName != null, nameof(callerName) + " != null");
 
             lock (this.objLock)
             {
                 return this.invocators.GetOrAdd(callerName, _ =>
                 {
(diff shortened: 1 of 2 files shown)


softworkz:
I had an AI running and working on this for almost 20h - just a few interruptions.

All-in-all it was pretty successful, but only because I had stictly forbidden to make any edits from its own "knowledge" and use a docs MCP server which I had previously fed with the Electron documentation for exactly one version (39.2). This allowed me to check occasionally, whether it had actually queried the information from there. Anyway, my indoctrination has apprently been extreme but successful. It affirmed in every 2nd sentence that all information iis solely taken from the docs server.

I had checked back with the original docs and didn't see any mistake. What could have happened that it missed some new members in the classes (due to the methodology via querying MCP for docs), but  I would consider the total result as pretty solid.

Finally, I have reviewed every single changed file and made a number of editis (which also led to the additional commits), also reverted some unnecessary changes in favor of compatibility.

softworkz:
I had also give special instructions regarding numeric types (doing an assessment whether integral or fractional when the docs to not state anything and build a list of properties where it's unclea).

The ultimate criteria for me were as to whether it would properly identify and change the type of the property which caused the issue that @markatosi had run into (decimal displayFrequecy but int in our definitions): https://github.com/ElectronNET/Electron.NET/issues/944

And it did.

softworkz:
### Summary

With the extended matrix tests, 10 new test failures have surfaced: https://github.com/ElectronNET/Electron.NET/actions/runs/19586773630/attempts/1#summary-56097317718

Two of them had turrned out to be a timeout issue on newer Electron versions on Windows (takes 1.5s).
It's fixed in this PR (commit "getPrinters: Increase timeout").

The remaining 8 failures were automatically fixed by the model update in this PR (without having been aiming at that).

Now these are the results:

image


But I'm going to remove the matrix-test commit because it belongs into [the other PR](https://github.com/ElectronNET/Electron.NET/pull/948).

FlorianRappl (approved):
This is awesome!

PR #951, opened 2025-11-24
adding macOS Intel runner

agracio:
Added `macos-15-intel` runner to test older macOS Intel systems.

Closes #951


softworkz:
When tests are failing so arbitrarily like this:

image


...then it's usually about timing and needs adding some delay between the calls made in those tests.

I re-ran the failing tests and now they're all green.

PR #962, opened 2025-12-06
Collection of test fixes

softworkz:


src/ElectronNET.API/API/WebContents.cs
```diff
@@ -388,7 +388,7 @@ public void SetAudioMuted(bool muted)
     /// Returns string - The user agent for this web page.
     /// </summary>
     /// <returns></returns>
-    public Task<string> GetUserAgentAsync() => InvokeAsync<string>();
+    public Task<string> GetUserAgentAsync() => InvokeAsyncWithTimeout<string>(3000);
```
FlorianRappl: Should we always use the underscore as a separator for digits > 1000 (1_000, 1_000_000, ...). To be consistent also with line 176. I'd actually prefer that (consistency + readability).

src/ElectronNET.IntegrationTests/Tests/BrowserWindowTests.cs
```diff
@@ -120,22 +135,28 @@ public async Task MenuBar_auto_hide_and_visibility()
...
+                // Trigger a navigation and wait for DOM ready so the renderer paints, which emits ready-to-show
+                var domReadyTcs = new TaskCompletionSource();
+                window.WebContents.OnDomReady += () => domReadyTcs.TrySetResult();
+                await Task.Delay(500);
+                await window.WebContents.LoadURLAsync("about:blank");
+                await domReadyTcs.Task;
 
-            // Typical usage is to show once ready
-            window.Show();
+                var completed = await Task.WhenAny(tcs.Task, Task.Delay(3000));
```
FlorianRappl: 3_000 ?

src/ElectronNET.IntegrationTests/Tests/BrowserWindowTests.cs
```diff
@@ -120,22 +135,28 @@ public async Task MenuBar_auto_hide_and_visibility()
...
+                completed.Should().Be(tcs.Task);
+            }
+            finally
+            {
+                window?.Destroy();
+            }
         }
 
         [Fact(Timeout = 20000)]
```
FlorianRappl: Not sure if we should just apply this refactoring / readability improvement to the whole codebase to have also the existing ones (20000 -> 20_000) updated. Thoughts?

src/ElectronNET.API/API/WebContents.cs
softworkz: Yea, I'm the one who introduced it and didn't stick to it.

But now I've added something different that I'm using in all project, let me know how you like that...

src/ElectronNET.IntegrationTests/Tests/BrowserWindowTests.cs
softworkz: see above

softworkz:
@agracio 

I was able to remove the CI skip for all the tests. When there are inconsistent results, it has been almost always a timing issue.

Yesterday, I've run the tests here 5 times in a row and all were always successful.

For the future, there's a new `IntegrationFactAttribute` which has a `SkipOnCI` parameter that you can use when it would be needed in some case.

agracio:
Sounds great, the more tests run on CI the better. It is always annoying to remove tests when they start failing randomly on certain CI agents and I am glad you got to the bottom of it.

softworkz:
The problem in this case was use of the MainWindow by multiple tests. There was one test which made it show and hide - and this changed the state and behavior of the window in subsequent tests, so other tests' results became dependent on the execution order (which is not guaranteed). Hence I changed a number of tests to create their own window to make them work independently.

Like the ZoomLevel test for example is  working with a hidden (never-been-shown) window on some platforms but on others it requires a window that has been shown at least once.

PR #965, opened 2025-12-07
Fix publishing documentation

softworkz:


AeonSake:
Just a few notes from my testing with publish profiles:
- `ProjectGuid` is not required
- `Configuration` is required IF used in the dynamic `PublishUrl` property; alternatively you could also replace `$(Configuration)` with `$(LastUsedBuildConfiguration)`

softworkz:
> Just a few notes from my testing with publish profiles:
> 
> * `ProjectGuid` is not required

Probably not, but I'm not sure about it. There is surely **some** reason why the publishing wizard adds it.


> * `Configuration` is required IF used in the dynamic `PublishUrl` property; alternatively you could also replace `$(Configuration)` with `$(LastUsedBuildConfiguration)`

This doesn't work when there's no `LastUsedBuildConfiguration` - e.g. when publishing a project for the first time or building on CI.

AeonSake:
> > * `Configuration` is required IF used in the dynamic `PublishUrl` property; alternatively you could also replace `$(Configuration)` with `$(LastUsedBuildConfiguration)`
> 
> This doesn't work when there's no `LastUsedBuildConfiguration` - e.g. when publishing a project for the first time or building on CI.

Fair point; my point is that the publish profile itself doesn't work if no `Configuration` property is set _in the profile itself_ for some reason (also the publishing path will be incorrect). Either that, or the `LastUsedBuildConfiguration` property must be set. Also, the interface in Visual Studio gets somewhat confused if the `LastUsedBuildConfiguration` and `LastUsedPlatform` properties are missing (the UI shows "Unknown" for those fields), but it works regardless.

softworkz:
You are absolutely right: Configuration and Platform must be present. 

I've updated this PR, thanks!

softworkz:
This is included in the other PR now: https://github.com/ElectronNET/Electron.NET/pull/966

PR #968, opened 2025-12-07
feat: Add ElectronHostHook sample application (#967)

### LINKED ISSUE #967: Ideas for Sample Applications
If somebody would be interested in contributing app samples, here are some ideas:

- **How to inject a Preload Script** 
   e.g. for setting up IPC messaging between web content and .net ap
- **Build Customization via electron-builder hook scripts**
  like `beforePack`, `afterPack`, `artifactBuildStarted`, `artifactBuildCompleted`
  This allows for example to manipulate or replace the `*.desktop` file for Linux apps
- **A comprehensive sample for electron-builder.json**
  That's something you can hardly find anywhere...
- **Multi-Project Example**
  How you can have a base project with Electron.NET code that is shared by multiple Electron.NET apps
- **ElectronHostHook Sample**
  A sample project, showing just ElectronHostHook without any distractions
- **App Icons Sample**
  A working sample with icons for all 3 platforms (Win, Linux, Mac)

(for most of these, I can provide something based on which a sample can be created)

> [!NOTE]
>
> A good sample should minimal in general. While it can be more complex in the topic that it demonstrates, everything aound that is unrelated must be as minimal as possible and should not use any other patterns - which might be elegant and useful - but they never are for the one who is interested in the topic of the sample - such things are just a distraction.
> 
> It's better to have many samples, each of which demonstrating one specific featuie or usage, than a huge all-in-one sample where a new user would have no clue where one feature ends and another one starts in the code.

 

 _Originally posted by @softworkz in [#947](https://github.com/ElectronNET/Electron.NET/issues/947#issuecomment-3620429486)_

adityashirsatrao007:
Closes #967

## Changes
- Added new sample project \ElectronNET.Samples.ElectronHostHook\ demonstrating bidirectional communication using \ElectronHostHook\ logic.
- Implemented \ElectronHostHook/index.ts\ to handle 'ping' event and return 'pong'.
- Implemented HomeController to call the hook from C#.
- Added the new sample to \ElectronNET.sln\.

## Details
This addresses the request for a focused 'ElectronHostHook Sample' in the linked issue. The sample is minimal and self-contained, showing only the hook mechanism.

### DIFF
--- a/src/ElectronNET.Samples.ElectronHostHook/ElectronHostHook/.gitignore
+++ b/src/ElectronNET.Samples.ElectronHostHook/ElectronHostHook/.gitignore
@@ -1,0 +1,3 @@
+node_modules
+*.js
+*.js.map
--- a/src/ElectronNET.Samples.ElectronHostHook/ElectronHostHook/connector.ts
+++ b/src/ElectronNET.Samples.ElectronHostHook/ElectronHostHook/connector.ts
@@ -7,12 +7,15 @@ export class Connector {
     on(key: string, javaScriptCode: Function): void {
         this.socket.on(key, (...args: any[]) => {
             const id: string = args.pop();
-            const done = (result: any) => {
-                this.socket.emit(id, result);
-            };
-
-            args = [...args, done];
-            javaScriptCode(...args);
+            try {
+                javaScriptCode(...args, (data) => {
+                    if (data) {
+                        this.socket.emit(`${key}Complete${id}`, data);
+                    }
+                });
+            } catch (error) {
+                this.socket.emit(`${key}Error${id}`, `Host Hook Exception`, error);
+            }
         });
     }
 }
--- a/src/ElectronNET.Samples.ElectronHostHook/ElectronHostHook/package.json
+++ b/src/ElectronNET.Samples.ElectronHostHook/ElectronHostHook/package.json
@@ -1,0 +1,12 @@
+{
+  "name": "electron-host-hook",
+  "version": "1.0.0",
+  "description": "Connector for Electron.NET projects.",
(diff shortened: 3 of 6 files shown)


softworkz:
@adityashirsatrao007 - Wow, cool! That was fast (after I posted the issue at least)..

I think we need to determine a number of conventions for how to include the samples in the repo.

The existing test apps are meant for development. They have project files that allow them to consume all of Electron.NET directly, bypassing nuget packages (unless modified).
The actual samples should not be like that. They should be minimal und use the nuget packages just like everybody is using them.  As such, they also shouldn't be part of the development solution and they shouldn't be in the src folder but rather in their own folder - obviously `./samples` - I'd say.

The folder name should be the subject without prefix, i.e. in this case `ElectronHostHook`.

As a 3-part root namespace is a bit ugly, I'd suggest that all samples should have two parts

1. `ElectronNETSample` or `ElectronNetSample`
2. The subject - e.g. `ElectronHostHook`

So, the full paths would be:

`samples/ElectronHostHook/ElectronNetSample.ElectronHostHook.sln`
`samples/ElectronHostHook/ElectronNetSample.ElectronHostHook.csproj`

But please don't take my word for it and don't make any changes yet. 
We want to see what @FlorianRappl is thinking about it first.

softworkz:
@adityashirsatrao007 - Thanks for implementing all the Copilot suggestions I've marked them as resolved.

We're just waiting for input from @FlorianRappl now.

FlorianRappl:
I think we can stay with the 3-part solution (yes it's ugly, but it has the arguably "correct" hierarchy).

Let's merge it in like that. Thanks a lot for the effort! 🚀

PR #974, opened 2025-12-10
Drop .NET 6.0, default to .NET 10.0 and update dependencies

davidroth:
Updated all projects to target .NET 10.0, removing support for .NET 6.0. (Out of support since 1 year) 
.NET 6.0 is out of support since over 1 year, it makes no sense to build for it and is annoying if one does not have the 6.0 SDK installed.

Upgraded key dependencies, including:
- `Microsoft.Build` and `Microsoft.Build.Tasks.Core` to 18.0.2.
- `Nuke.Common` to 10.1.0.
- `System.Drawing.Common` to 10.0.1.
- `System.Text.Json` to 10.0.1.

### DIFF
--- a/src/ElectronNET.API/ElectronNET.API.csproj
+++ b/src/ElectronNET.API/ElectronNET.API.csproj
@@ -3,7 +3,7 @@
   <Import Project="..\common.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>net6.0;net8.0;net10.0</TargetFrameworks>
+    <TargetFrameworks>net8.0;net10.0</TargetFrameworks>
     <PackageOutputPath>..\..\artifacts</PackageOutputPath>
     <PackageId>$(PackageNamePrefix).API</PackageId>
     <Title>$(PackageId)</Title>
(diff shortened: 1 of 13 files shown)


src/ElectronNET.API/ElectronNET.API.csproj
```diff
@@ -28,8 +28,8 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
     </PackageReference>
     <PackageReference Include="SocketIOClient" Version="3.1.2" />
-    <PackageReference Include="System.Drawing.Common" Version="8.0.16" />
-    <PackageReference Include="System.Text.Json" Version="8.0.5" />
+    <PackageReference Include="System.Drawing.Common" Version="10.0.1" />
```
FlorianRappl, line 31: How would that work if you only have .NET 8 ? I think we need to reference the lower ones here.

softworkz: Those numbers mean nothing. It is required to look up for every single one, wether it still supports net6 and net8.

softworkz: But System.Text.Json cannot be dropped. It's not included in net6.

softworkz: System.Text.Json gets transitively pulled in by SocketIOClient, but when it's removed, it falls down to a vulnerable version on net10.

FlorianRappl:
Must be rebased to `develop`.

src/ElectronNET.API/ElectronNET.API.csproj
davidroth: Yep, each of them still has a 8.0 target.
I explicitly dropped 6.0 in this MR (see description)

davidroth:
rebase done

PR #992, opened 2025-12-16
Try fix job re-run

softworkz:


softworkz:
Please merge over to main.

It's only effective from the default branch (that's why it was never tested).

Thanks

PR #994, opened 2025-12-18
Fix single instance broken

davidroth:
I was debugging why my electron app does not respect the singleInstance property and found out that commit 688d6d83cc0d8f70ced6821a1bd0a5e12472e26f regressed the single instance behaviour.

Problem: Commit 688d6d83cc0d8f70ced6821a1bd0a5e12472e26f changed the check to == 'yes' but the configuration uses `true`.

### DIFF
--- a/src/ElectronNET.API/Runtime/StartupManager.cs
+++ b/src/ElectronNET.API/Runtime/StartupManager.cs
@@ -167,11 +167,11 @@
 
                 if (isSingleInstance?.Length > 0 && bool.TryParse(isSingleInstance, out var isSingleInstanceActive) && isSingleInstanceActive)
                 {
-                    buildInfo.ElectronSingleInstance = "yes";
+                    buildInfo.ElectronSingleInstance = "true";
                 }
                 else
                 {
-                    buildInfo.ElectronSingleInstance = "no";
+                    buildInfo.ElectronSingleInstance = "false";
                 }
 
                 if (httpPort?.Length > 0 && int.TryParse(httpPort, out var port))
--- a/src/ElectronNET.Host/main.js
+++ b/src/ElectronNET.Host/main.js
@@ -93,7 +93,7 @@ app.on('will-finish-launching', () => {
 
 const manifestJsonFile = require(manifestJsonFilePath);
 
-if (manifestJsonFile.singleInstance === "yes") {
+if (manifestJsonFile.singleInstance === "true" || manifestJsonFile.singleInstance === "True") {
     const mainInstance = app.requestSingleInstanceLock();
     app.on('second-instance', (events, args = []) => {
         args.forEach((parameter) => {


davidroth:
@FlorianRappl I tested it with my changes and it works now (both enabling + disabling). Not exactly sure why you changed it to "yes" in the original commit -maybe you can clarify? Thanks 😀

Tested by running the following command in the published output dir:

Manipulate "package.json" and configure "singleInstance" to either "true" or "false"

Run app twice via:

`node node_modules/electron/cli.js main.js -unpackedelectron`


### Previously:

App always runs twice.

src/ElectronNET.Host/main.js
```diff
@@ -93,7 +93,7 @@ app.on('will-finish-launching', () => {
 
 const manifestJsonFile = require(manifestJsonFilePath);
 
-if (manifestJsonFile.singleInstance === "yes") {
+if (manifestJsonFile.singleInstance === "true" || manifestJsonFile.singleInstance === "True") {
```
davidroth, line 96: The Msbuild property, when changed is set to "True" (uppercase). The default value is "true" (lowercase):

Image

src/ElectronNET.API/Runtime/StartupManager.cs
```diff
@@ -167,11 +167,11 @@ private BuildInfo GatherBuildInfo()
 
                 if (isSingleInstance?.Length > 0 && bool.TryParse(isSingleInstance, out var isSingleInstanceActive) && isSingleInstanceActive)
                 {
-                    buildInfo.ElectronSingleInstance = "yes";
+                    buildInfo.ElectronSingleInstance = "true";
```
davidroth, line 170: @FlorianRappl For what is the buildInfo used ? I could not find any meaningful usages of `buildInfo.ElectronSingleInstance`

FlorianRappl: I don't understand the question (you already see that this is an object that is serialized and passed on - this object contains a bunch of properties).

Also, I think you confuse the property (True / False) with the provided string, which is handed over the Node.js counterpart.

src/ElectronNET.Host/main.js
FlorianRappl: This has nothing (directly) to do with the MSBuild property - we receive this in JS, which is based on the object created earlier.

The relation is purely indirect, i.e., the value set in the configuration is used to create the object that is used here.

FlorianRappl:
Yeah this I think is an artifact from the migration to pure .NET, i.e., no more serialization here. I think it might even make sense to remove the `BuildInfo` altogether. I don't see much value in this any more.

As far the issue - I think removal of the deserialization and comparing to `True` would be sufficient.

src/ElectronNET.Host/main.js
davidroth: I see something different in the source code.
The value is read from `manifestJsonFileName` which is `package.json` per default.

`ElectronNET.LateImport.targets`  file generates `package.json` based on `package.template.json`.

It uses the value from the TempalteProperties:

`<TemplateProperty Include="ElectronSingleInstance" Value="$(ElectronSingleInstance)" />`

Therfore the msbuild property directly gets serialized into package.json which is then used in main.js.

FlorianRappl: Yeah so the `BuildInfo` is not (almost) purely informational. As such I think it makes sense to

1. Remove the computation / deserialization of `ElectronSingleInstance` altogether
2. Keep the comparison on `True` (imho it's fine to keep it case-invariant, but feel free to be more strict here, too)

src/ElectronNET.API/Runtime/StartupManager.cs
FlorianRappl: Obsolete - see other comment (https://github.com/ElectronNET/Electron.NET/pull/994#discussion_r2631319775).

davidroth: I can't find any use for `buildInfo.ElectronSingleInstance` anywhere, nor any point where `buildInfo` is serialized or passed on.

Either I'm missing something, or this function isn't used.

softworkz:
The original mistake is in package.template.json:

 ```json
 "license": "$(License)",
  "executable": "$(TargetName)",
  "singleInstance": "$(ElectronSingleInstance)",
  "homepage": "$(ProjectUrl)",
  "splashscreen": {
    "imageFile": "$(ElectronSplashScreen)"
  },
```

The value should not be quoted.

And then in `ElectronNET.LateImport.targets`, this

```xml
<TemplateProperty Include="ElectronSingleInstance" Value="$(ElectronSingleInstance)" />
```

needs to be changed to this:

```xml
<TemplateProperty Include="ElectronSingleInstance" Value="$(ElectronSingleInstance.ToLower())" />
```

softworkz:
> Yeah this I think is an artifact from the migration to pure .NET, i.e., no more serialization here. I think it might even make sense to remove the `BuildInfo` altogether. I don't see much value in this any more.

BuildInfo is important. There are 3 internal usages:

image

For example, without it, the ElectronRuntime wouldn't know the name of the Electron executable to launch.

C# has no DEFINEs like C/C++, so it's not possible to get compile-time variables into the code, apart from the boolean-only definitions for conditional compilation.

Using assembly metadata attributes is an elegant replacement for this. The ElectronRuntime.BuildInfo unwraps those values, so that you don't need the ugly code everywhere for reading those attributes.

src/ElectronNET.API/Runtime/StartupManager.cs
softworkz: This is a public property that you can use in your application to know about the behavior (how it has been compiled).

softworkz:
@davidroth 

Thanks a lot for the PR and bringing this to attention. This PR is close, but actually, that value should be a boolean in the first place, that's why https://github.com/ElectronNET/Electron.NET/pull/994 is a better way from my point of view.

davidroth:
Closed in favour of https://github.com/ElectronNET/Electron.NET/pull/996

PR #1021, opened 2026-02-02
Secure Connection on Shared Machines

FlorianRappl:
This PR was done by third-party developers; I am just here to provide their changes for review.

The basic idea is outlined in the [Electron.NET SignalR Authentication Guide](docs/SignalR-Authentication-Guide.md).

In here, the system makes sure to use exclusively respond to a request when a cookie based on an initial auth token was provided. This way, even though the application behind is open to all users of the same machine, only the "correct" one is answering.

The background for this is that sometimes the desktop application is used by connecting to a remote machine effectively creating a user session on the machine. If this is done multiple times by different users then the same application is launched multiple times on this machine. In general, multi-instance mode works fine, but has of course some limitations such as requiring dedicated ports for each running app instance.


softworkz:
Sorry, but this is insane. I have no time to review this.

softworkz:
> Sorry, but this is insane. I have no time to review this.

There hasn't even any effort been made to make this reviewable. This is full of formatting, eol and other stray changes.

FlorianRappl:
> This is full of formatting, eol and other stray changes.

I don't agree on this assessment, but I agree that there are indeed some changes beyond the PR goal. I will try to merge those or remove them to clean this up.

src/ElectronNET.AspNet/Runtime/Services/AspNetLifetimeAdapter.cs
```diff
@@ -1,27 +1,28 @@
...
+using ElectronNET.Runtime.Services;
+
+using Microsoft.Extensions.Hosting;
+
+using System.Threading.Tasks;
+
+namespace ElectronNET.AspNet.Runtime;
+
+internal class AspNetLifetimeAdapter : LifetimeServiceBase
 {
```
softworkz: Usings location changed.
Namespace changed to file scope
EOF changed

src/ElectronNET.AspNet/Runtime/Services/AspNetLifetimeAdapter.cs
```diff
@@ -1,27 +1,28 @@
...
 
-        protected override Task StopCore()
-        {
-            this.lifetimeService.StopApplication();
-            return Task.CompletedTask;
-        }
+    protected override Task StopCore()
+    {
+        this.lifetimeService.StopApplication();
+        return Task.CompletedTask;
```
softworkz: This is shown as change even though there is none

src/ElectronNET.Host/api/app.ts
```diff
@@ -1,313 +1,338 @@
-import { RelaunchOptions, LoginItemSettingsOptions, Settings, AboutPanelOptionsOptions } from "electron";
+import {
+  RelaunchOptions,
+  LoginItemSettingsOptions,
+  Settings,
+  AboutPanelOptionsOptions,
+} from "electron";
```
softworkz, line 7: Formatting change

src/ElectronNET.Host/api/app.ts
```diff
@@ -1,313 +1,338 @@
...
+  });
+
+  socket.on("appSetAppLogsPath", (path) => {
+    app.setAppLogsPath(path);
+  });
+
+  socket.on("appGetPath", (name) => {
+    const path = app.getPath(name);
+    electronSocket.emit("appGetPathCompleted", path);
+  });
```
softworkz, line 127: Shown as change even though there is none like this.

src/ElectronNET.Host/api/autoUpdater.ts
```diff
@@ -1,143 +1,192 @@
...
-    socket.on('autoUpdater-getFeedURL', async (guid) => {
-        const feedUrl = await autoUpdater.getFeedURL();
-        electronSocket.emit('autoUpdater-getFeedURL-completed' + guid, feedUrl || '');
+  socket.on("register-autoUpdater-error", (id) => {
+    autoUpdater.on("error", (error) => {
+      electronSocket.emit("autoUpdater-error" + id, error.message);
     });
+  });
+
+  socket.on("register-autoUpdater-checking-for-update", (id) => {
```
softworkz, line 15: Same again. This is not reviewable.

src/ElectronNET.Host/api/browserView.ts
```diff
@@ -1,74 +1,82 @@
...
+import { Socket } from "net";
+import { BrowserView } from "electron";
 
-const browserViewApi = (socket: Socket) => {
-    electronSocket = socket;
+const browserViews: BrowserView[] = (global["browserViews"] =
+  global["browserViews"] || []) as BrowserView[];
+const proxyToCredentialsMap: { [proxy: string]: string } = (global[
+  "proxyToCredentialsMap"
+] = global["proxyToCredentialsMap"] || []) as { [proxy: string]: string };
```
softworkz: Formatting changes

src/ElectronNET.Host/api/browserWindows.ts
```diff
@@ -141,7 +144,11 @@ export = (socket: Socket, app: Electron.App) => {
 
   socket.on("register-browserWindow-bounds-changed", (id) => {
     const window = getWindowById(id);
-    const cb = () => electronSocket.emit("browserWindow-bounds-changed" + id, window.getBounds());
+    const cb = () =>
+      electronSocket.emit(
+        "browserWindow-bounds-changed" + id,
+        window.getBounds(),
+      );
```
softworkz, line 154: Formatting change

src/ElectronNET.Host/api/browserWindows.ts
```diff
@@ -231,7 +238,7 @@ export = (socket: Socket, app: Electron.App) => {
         __dirname,
         "..",
         "scripts",
-        "blazor-preload.js"
+        "blazor-preload.js",
```
softworkz, line 244: Comma change

src/ElectronNET.Host/api/clipboard.ts
```diff
@@ -1,84 +1,87 @@
...
+  });
+
+  socket.on("clipboard-availableFormats", (type) => {
+    const formats = clipboard.availableFormats(type);
+    electronSocket.emit("clipboard-availableFormats-completed", formats);
+  });
+
+  socket.on("clipboard-write", (data, type) => {
+    clipboard.write(data, type);
+  });
```
softworkz, line 64: Again, the full block shown as change

src/ElectronNET.Host/api/ipc.ts
```diff
@@ -1,81 +1,90 @@
...
+    const window = BrowserWindow.fromId(browserWindow.id);
 
-        if (window) {
-            window.webContents.send(channel, ...data);
-        }
-    });
+    if (window) {
+      window.webContents.send(channel, ...data);
+    }
+  });
```
softworkz, line 42: Whitespace change

src/ElectronNET.Host/api/browserWindows.ts
FlorianRappl: I think you are focusing on the wrong area.

Most of the changes in the host are formatting changes (formatted as it actually should be - but as written these should be merged already into develop and therefore not be in this PR).

softworkz:
> > This is full of formatting, eol and other stray changes.
> 
> I don't agree on this assessment, but I agree that there are indeed some changes beyond the PR goal. I will try to merge those or remove them to clean this up.

Did you ever look at the GitHub diff before disagreeing?

FlorianRappl:
> Did you ever look at the GitHub diff before disagreeing?

Yes. Most of the formatting changes (like 90%) are in the Host section. And that is the one that should be excluded (these are Prettier changes and the resulting formatting is actually as it should be; the previous formatting is just awful).

softworkz:
Just to be clear: The comments I made were just examples - like I said, the whole PR is full of those kinds of changes.

FlorianRappl:
> Just to be clear: The comments I made were just examples - like I said, the whole PR is full of those kinds of changes.

Do you know how a tree view works? Because you can just collapse the Host directory. I am sorry for not being direct here; I think I wrote that there are some changes which should not be part of the PR. But that was not the point of the request. The point was the general change.

Now, again this is agreeable a mess. But "full of" would imply that all of the .NET code is also just there due to formatting changes, which certainly is not the case. So I am not sure we have the same base line here.

Where I am coming from is a conceptual review - mostly of the .NET code (SignalR integration) and the design documents. This is what I am looking at, too - I was interested in your opinion on this.

softworkz:
When there are no changes to the Host files, why are they in this PR, then?

In all PRs I have made, I have cleanly separated formatting changes from functional changes. Not from a nitpicking perspective, but for the simple reason that you cannot properly review code changes when there are hundreds of changes which aren't actual changes. I'm not willing - nor do I have the time - to look through 128 changed files. 

From a high-level perspective, my verdict is that a PR which needs changes in 128 files isn't worth considering at all at this point (where we are just about to stabilize things).

softworkz:
> This PR was done by third-party developers;

Developers named Claude, Copilot or GPT, I assume...

FlorianRappl:
> When there are no changes to the Host files, why are they in this PR, then?

I did not write "no changes". But nevermind (I think only 2 files contained changes - most importantly `main.js`).

> In all PRs I have made, I have cleanly separated formatting changes from functional changes. Not from a nitpicking perspective, but for the simple reason that you cannot properly review code changes when there are hundreds of changes which aren't actual changes. I'm not willing - nor do I have the time - to look through 128 changed files.

As mentioned - fair enough. Also, of course its good to separate. The problem with the mixing of the formatting happened on my side (as the commit history shows); hence my comment to ignore these for now. That would still be around a hundred files, so as mentioned it's reasonable that this is too much. I never wrote anything against it.

> From a high-level perspective, my verdict is that a PR which needs changes in 128 files isn't worth considering at all at this point (where we are just about to stabilize things).

Fair enough, too. Right now this is not about to rush it, but (as pointed out in the other comment that still seems to be ignored) about the concept. Maybe I should have created this as a draft, but I always dislike draft PRs especially when I want to see what our automation shows for it.

> Developers named Claude, Copilot or GPT, I assume...

True - that's also what was communicated to me that the bulk work has been done by coding agents. But I personally do not care what tools the people use, I only followed up on devs behind the proposal and their request to have a look and see if there is something useful for us.

While I understand your position I do not agree with your attitude. Maybe the writing does not help - but there is really no need to act like this. You have no time? This is fine. You think this is too much / not being brought in using the right format? Fair enough. But please keep the attitude aside.

softworkz:
> While I understand your position I do not agree with your attitude. Maybe the writing does not help - but there is really no need to act like this. You have no time? This is fine. You think this is too much / not being brought in using the right format? Fair enough. But please keep the attitude aside.

I'm slightly annoyed, yes. Wouldn't you be when somebody would throw gigantic AI-generated mess at you, asking to review it?

I have nothing against AI, I'm using it myself, yet I'm always transparent and mention it. From a professional side I also need to say that those changes cannot be trusted. When AI is used for coding in a way that it rewrites whole files (from "memory"), which has obviously been the case here, then there's always a high risk for mistakes, because at some point information is falling out of the context window and LLMs are not aware of that. Instead, they are producing something similar but no longer accurate.
This - in combination with all the formatting changes plus the large amount of files is a recipe for failure, because nobody wants to go through so many changes letter-by-letter.
State of the art meanwhile is to have the models do targeted edits only. And before that, the recipe was to be fully intolerant regarding any stray changes and to insist on preserving everything unrelated.

softworkz:
Just another note - maybe there was a misunderstanding regarding my saying "I have no time". I meant and mean it **literally**.
Because when I really had to deal with this, I would start by eliminating all the non-changes - partially by tooling and the rest by hand. I've done this a number of times where I really had no choice. And hence I know that this is a monkey-job and which requires a lot of work.

FlorianRappl:
> I'm slightly annoyed, yes. Wouldn't you be when somebody would throw gigantic AI-generated mess at you, asking to review it?

That's literally what happened here. Hence my request for you to look at the concept. Again, I take the blame that the formatting of the TS came into this one and that I should have been more direct w.r.t. the objective. However, I did write it a couple of times and you seem to ignore that point.

softworkz (changes requested):
> That's literally what happened here. Hence my request for you to look at the concept. Again, I take the blame that the formatting of the TS came into this one and that I should have been more direct w.r.t. the objective. However, I did write it a couple of times and you seem to ignore that point.

Yes, because with this insane diff, I'm not willing to dig through it trying to understand any concept.

Even when ignoring the Host project, these changes are done in a way that someone may do when using the code for oneself alone, but this is not a suitable contribution to an open-source project. Change all logging - WTH?
These changes are invasive without taking care about the project as-is,, which is evident from a range of indicators. There's no effort visible for making a good contribution to the project, and in this regard, I see no reason why I should take effort then, in analyzing the concept. 


This is a clear rejection from my side, it's not acceptable in this form
(neither for me to look at it any further)

src/ElectronNET.API/Runtime/Helpers/DebuggerHelper.cs
```diff
@@ -0,0 +1,18 @@
...
+    internal static class DebuggerHelper
+    {
+        /// <summary>
+        /// Gets whether a debugger is attached. This value is cached for performance.
+        /// </summary>
+        public static bool IsAttached => _isAttached.Value;
+
+        private static readonly Lazy<bool> _isAttached = new Lazy<bool>(() => Debugger.IsAttached);
+    }
+}
```
softworkz: I'm not aware that Debugger.IsAttached would be a very expensive call. 
Even less that it would be so expensive that it's worth all this extra code to reduce invocations from 4 to 1. 
I might be wrong, but either way: this is unrelated and should be in a separate PR (if at all).

src/ElectronNET.API/Runtime/Helpers/UnpackagedDetector.cs
```diff
@@ -63,13 +63,16 @@ public static bool CheckIsUnpackaged()
...
+            var dir = new DirectoryInfo(baseDir);
+            
             if (dir.Name == "bin" && dir.Parent?.Name == "resources")
             {
                 return false;
             }
 
-            if (dir.GetDirectories().Any(e => e.Name == ".electron"))
+            // Faster: Direct path check instead of directory enumeration
+            if (Directory.Exists(Path.Combine(baseDir, ".electron")))
```
softworkz: Faster - yes, but when there are 1000 invocations of this method - here we got a signle one only...

src/ElectronNET.API/Runtime/StartupManager.cs
```diff
@@ -25,9 +25,7 @@ public void Initialize()
             this.CollectProcessData();
             this.SetElectronExecutable();
 
-
             ElectronNetRuntime.StartupMethod = this.DetectAppTypeAndStartup();
-            Console.WriteLine((string)("Evaluated StartupMethod: " + ElectronNetRuntime.StartupMethod));
```
softworkz, line 30: This one seems to be gotten los

src/ElectronNET.AspNet/API/WebHostBuilderExtensions.cs
```diff
@@ -66,23 +67,31 @@ public static IWebHostBuilder UseElectron(this IWebHostBuilder builder, string[]
             // work as expected, see issue #952
             Environment.SetEnvironmentVariable("ELECTRON_RUN_AS_NODE", null);
 
-            var webPort = PortHelper.GetFreePort(ElectronNetRuntime.AspNetWebPort ?? ElectronNetRuntime.DefaultWebPort);
+            // For SignalR modes, use port 0 for dynamic port assignment
+            var usePort0 = ElectronNetRuntime.StartupMethod == StartupMethod.PackagedDotnetFirstSignalR ||
+                          ElectronNetRuntime.StartupMethod == StartupMethod.UnpackedDotnetFirstSignalR;
+
+            var webPort = usePort0 ? 0 : PortHelper.GetFreePort(ElectronNetRuntime.AspNetWebPort ?? ElectronNetRuntime.DefaultWebPort);
```
softworkz: A switch might be better readable

src/ElectronNET.AspNet/API/WebHostBuilderExtensions.cs
```diff
@@ -97,6 +106,17 @@ public static IWebHostBuilder UseElectron(this IWebHostBuilder builder, string[]
                     case StartupMethod.UnpackedDotnetFirst:
                         services.AddSingleton<IElectronNetRuntimeController, RuntimeControllerAspNetDotnetFirst>();
                         break;
+                    case StartupMethod.PackagedDotnetFirstSignalR:
+                    case StartupMethod.UnpackedDotnetFirstSignalR:
```
softworkz: Since SignalR is always and only dotnet-first, it might add clarity to just use 

`PackagedSignalR`
`UnpackedSignalR`

src/ElectronNET.AspNet/Bridge/SignalRFacade.cs
```diff
@@ -0,0 +1,232 @@
...
+    /// for compatibility with existing Electron API code.
+    /// 
+    /// Key implementation details:
+    /// - Uses IHubContext to send events to Electron via 'event' hub method
+    /// - Receives events from Electron via ElectronHub.ElectronEvent() method
+    /// - Includes ConvertToType&lt;T&gt; helper to handle JsonElement and numeric type conversions
+    /// - Event args are passed as arrays to match SignalR serialization behavior
+    /// - Connection ID is set by ElectronHub when Electron client connects
+    /// </summary>
+    internal class SignalRFacade : IFacade
```
softworkz: - As we need an interface here, it might a good chance to get rid of the unhandy work "Facade". 

How about naming it like this:

- IFacade => ISocketConnection
- SocketIoFacade  => SocketIOConnection
- SignalRFacade  => SignalRConnection

src/ElectronNET.Host/logger.js
```diff
@@ -0,0 +1,199 @@
...
+
+module.exports = {
+  logger,
+  safeConsole,
+  LogLevel,
+  setLogLevel,
+  getLogLevel,
+  getEnvironment,
+  shouldLog,
+};
```
softworkz: This looks unrelated to this PR...no?

Should be separated.

src/ElectronNET.Samples.ElectronHostHook/Properties/electron-builder.json
```diff
@@ -0,0 +1,19 @@
...
+  },
+  "win": {
+    "target": [
+      {
+        "target": "portable",
+        "arch": "x64"
+      }
+    ]
+  }
+}
```
softworkz, line 19: Unrelated

src/ElectronNET/build/package.template.json
```diff
@@ -32,6 +32,7 @@
     "start": "tsc -p ."
   },
   "dependencies": {
+    "@microsoft/signalr": "^8.0.7",
```
softworkz: Non-Blazor apps should not get this package

softworkz (commented):
Thanks a lot for cleaning this up - even though it's incomplete and a lot of the ugly stuff is just hidden under the carpet - it's a lot better to look at now, and I see clearly (what I wasn't able to see before) how we can pull this out of the messiness corner:

The determination between Regular and Signalr should be a 

## Build-Time Decision

### Rough Outline

- We add new checkbox in the project designer: **Use Signaler**
  (i.e.: MSBuild property)
- It's only visible when the project is an ASP.NET project
  (+ build error when set with a non-asp.net project)
- This build property gets into the project via assembly attributes (like we already have some)
- For certain files, we have two versions: One for SignalR, one for normal
  - main.js, possibly some of the host files - just where it makes sense  
  - Maybe also for the package.json template (otherwise, we need to change it depending on "Use Signalr", so that SignalR dependencies are only included when "Use Signalr" is enabled.
  - vice-versa: no SocketIO deps when SignaLR is used
- In turn, it's not an environment vaiable to decide about  whether it's signalr or not 
- Unsupported cases should error out clearly to avoid confusion

### Benefits


- Better isolation between these modes of operation
- Higher clarity, less confusion
- Less error-prone, more robust
- Changes can be made to one, with less risk of affecting the other
- etc..

I haven't looked into every detail yet, but I hope this gives the orientation you were asking for. If you want to go for it, there's quite a bit of work to do.
I can help on the MSBuild side, if you wish.

Best,
sw

src/ElectronNET.API/Runtime/StartupManager.cs
FlorianRappl: Thanks - re-added.

src/ElectronNET.API/Runtime/Helpers/UnpackagedDetector.cs
FlorianRappl: The comment is misleading. I removed it (I think the code variant is fine, but the label "faster" is actually irrelevant due to the usage).

src/ElectronNET.API/Runtime/Helpers/DebuggerHelper.cs
FlorianRappl: You are right here - it definitely is not an expensive call. I removed that class and all usages.

src/ElectronNET.Host/logger.js
FlorianRappl: Yeah I'd like to remove this one. I will keep it for the moment to ease debugging / see what outputs really help but generally we will not introduce it here.

src/ElectronNET.AspNet/Bridge/SignalRFacade.cs
FlorianRappl: Sounds good to me. Before I change that I'd like to discuss what the future of SignalR is here. Me personally, I am not a big fan and for me SocketIO would be sufficient (but, of course, I see why SignalR might make more sense / at least for ASP.NET / Blazor).

src/ElectronNET.API/Runtime/Helpers/UnpackagedDetector.cs
softworkz: The change is valid, it just doesn't belong into a PR like this. 
Now, I don't want to be nitpicking, when there's one such change - okay - yet it was the high volume of changes of that kind which made me say that it's "full of stray changes".

src/ElectronNET.AspNet/Bridge/SignalRFacade.cs
softworkz: > Me personally, I am not a big fan and for me SocketIO would be sufficient

I see it the same.

> I'd like to discuss what the future of SignalR is here

Yup, I'll follow up in the main comments stream.

src/ElectronNET.Host/logger.js
softworkz: I can't see where it's used,  When there's a purpose, it's fine.

Generally, it would be great when we would be able to surface any errors on the JS side back to the .net API.

src/ElectronNET.Samples.ElectronHostHook/Properties/electron-builder.json
softworkz: I know it is missing in the sample. A separate PR (or just a commit to develop) would be good.

softworkz:
> I am not a big fan and for me SocketIO would be sufficient (but, of course, I see why SignalR might make more sense / at least for ASP.NET / Blazor).

> I'd like to discuss what the future of SignalR is here

I share those feelings. I understand the desire of having a kind of auth pattern for socket communication.

Can you lay out the advantages that using SignalR would provide besides the authentication - which I presume would also be doable via SocketIO?

FlorianRappl:
> Can you lay out the advantages that using SignalR would provide besides the authentication - which I presume would also be doable via SocketIO?

Personally, I don't see any. From my original conversation (via mail) with the original authors (or vibe coders if you prefer that label) of the PR they responded to my rejection of SignalR like this:

> Well… I understand your point about not using SignalR, but for the end-user who will be building a Blazor Server app with an Electron front-end, they already pay the price for SignalR on the backend. And implementing the node API over SignalR is almost a piece of cake 😉. But yes, I assume that WebSockets would also work, but require (for us) significantly more work.

As the connection would anyway not be used for the Blazor app itself I think just having SocketIO would be better. Maybe we can still keep the facade / interface (but rename it as you suggested). Generally, I think the mechanism itself (just start with some random token - make sure that both sides know about it - once agreed use it for protecting the line) can be done with SocketIO, too.

The mechanism of having (at build-time as you suggested) select the mode for the published app is also beneficial. This way, one can decide what side should be the carrier. To simplify things we could allow that token mechanism to only work if .NET is the initial runtime.

src/ElectronNET.Samples.ElectronHostHook/Properties/electron-builder.json
FlorianRappl: Yeah I will extract that out - just waiting to finish my thoughts for next steps on this one (i.e., how to proceed w.r.t. SignalR) before putting the PR into its final state.

src/ElectronNET.Host/logger.js
FlorianRappl: Most errors could be transported but as we communicate via the socket we would not be able to see any errors w.r.t. the socket connection.

src/ElectronNET.API/Runtime/Helpers/UnpackagedDetector.cs
FlorianRappl: Agreed.

softworkz:
> Personally, I don't see any. From my original conversation (via mail) with the original authors (or vibe coders if you prefer that label) of the PR they responded to my rejection of SignalR like this:
> 
> > Well… I understand your point about not using SignalR, but for the end-user who will be building a Blazor Server app with an Electron front-end, they already pay the price for SignalR on the backend. And implementing the node API over SignalR is almost a piece of cake 😉. But yes, I assume that WebSockets would also work, but require (for us) significantly more work.

I can't quite follow. I mean, the "auth" part needs to be implemented either here or there, it's the same in both cases. 
So, where's the "more work"? Getting SignalR into the game is more work. Using SocketIO doesn't require additional work.

Of course I (and probably 'we') are biased by not using SignalR ourselves, and we might think differently if we would. But at the end it's still us, being the ones who will have to support it, going forward. And that's my primary pain point about it - when someone comes and reports an issue about it, what should we do? Become SignalR experts? Or just shrug our shoulders?

If the PR would have at least been submitted by a SignalR expert who stands behind it, knows all about it, using it in their own product and would be available for assisting here with issues about it for a foreseeable amount of time - that would be a lot more convincing than the situation we're facing here.

I'm not fully decided, but I'm leaning towards being egoistic, rather not taking the burden and additional technical debt. I mean - why should we?

epsitec:
The code I submitted to Florian may give the impression that it comes from one of these too many, too enthusiastic vibe-coders. I believe I don't belong to that crowd, even if I totally agree with @softworkz initial evaluation: the submitted code stinks and was by no means intended to be submitted as a PR, without heavy rework/rewrite/checking/formatting/reviewing. It was more a proof of concept of the ideas I tried to express, and a means to verify that the architecture would work. That's why I sent it to Florian first, so that he can tidy things up (I'd love to do that myself but I have no available time, so we are subcontracting the work).

I am awfully sorry if the whole experiment gave the impression that I was being rude and didn't give a s**t about the practices which are in place for ElectronNET. More so with my very long past as an Open Source contributor...

I understand your reluctance to integrate SignalR into the equation (Florian had the same reaction). This adds one more moving piece to the system, and it is higher level than plain websockets (or socketio for that matter). However, apps which use a Blazor Server backend already have everything in place: SignalR is what drives Blazor's communication between the ASP.NET Core layer and the browser.

My intent is very clear: I need to get rid of the port used for `socketio` and have **only** ASP.NET Core pick and open a port (hence 127.0.0.1:0 for Kestrel), then have Electron be a placid system which connects to my Kestrel instance. This makes the whole system much simpler from our point of view, even though it means using SignalR in a somewhat unusual way, where the server (.NET) drives the client (main.js).

So SignalR elegantly solves our needs. And I will stand behind the choice. And invest what is required to have SignalR supported in the long run.

softworkz:
> I am awfully sorry if the whole experiment gave the impression that I was being rude and didn't give a s**t about the practices which are in place for ElectronNET. More so with my very long past as an Open Source contributor...

I'm also contributing to FFmpeg (member of the GA) and when you do similar there, you may get responses where you might just want to turn around and never come back :-)
But at the core, it's still very valid: why should I spend my time for digging through a load of sh..t when the other one is (seemingly) not willing to spend their own time for submitting a proper PR? Because when there is no visible dedication and passion behind, it leads you to a range of conclusions, every single one shouting out loud: "run away, you're wasting your time here".

Apologies accepted though. 


> My intent is very clear: I need to get rid of the port used for `socketio` and have **only** ASP.NET Core pick and open a port (hence 127.0.0.1:0 for Kestrel), then have Electron be a placid system which connects to my Kestrel instance. This makes the whole system much simpler from our point of view, even though it means using SignalR in a somewhat unusual way, where the server (.NET) drives the client (main.js).
> 
> So SignalR elegantly solves our needs. And I will stand behind the choice. And invest what is required to have SignalR supported in the long run.

Let me be honest with you: My intent and motivation has always been very clear as well. I'm not a hobbyist developer who is contributing to an open-source project for fun. I made that contribution for Electron.NET Core out of strategical needs. For once, I needed to be able to debug with little delay and to work and debug a Linux application from within Visual Studio on Windows. Also Electron.NET was about to die silently and slowly and it needed some boost in order to have a future (luckily, the changes are able to achieve all at once) - because nobody wants to go all in for a dead horse.

At the bottom line, there's no reason for me to deal with SignalR support in Electron.NET - it makes everything just more difficult and I need to take care that there won't be any changes that would regress our usage of it - so for me it's better when this doesn't get into Electron.NET.
Electron.NET is open-source with a very permissive license, so essentially you can already do what you want - this doesn't depend on what might get merged into Electron.NET and what not. Why do people still want to get their features into open-source projects? Well - because they are expecting that it keeps getting maintained there - done by others - so that "their feature" is always up-to-date and gets fixes and improvements - which are happening inevitably over time.

So,, as far as I'm concerned, I'm strongly leaning towards the "no"-side regarding SignalR integration.
But I'm leaving it up to Florian. If he wants to go for it, then I'll go with him.

FlorianRappl:
Thanks for the info @epsitec - I am still not sure why you insist on SignalR. From your perspective there should not be a difference. Sure, you will use SignalR in your code (which is fine) - but whatever the other connection (that you don't own nor can attach to) uses should be irrelevant.

As socketio will be part of the application anyway why have two libs? Sure I can understand your position that ASP.NET Core / Blazor anyway uses it, so if socketio would not be there it would be just this, but then again it would be strongly tied to ASP.NET Core.

So my position is that we can progress on this one (i.e., I will continue on it) by removing SignalR and have the secure connection done via socketio - same principles would apply here, just more simplified code imho.

epsitec:
@FlorianRappl this might be caused by my misunderstanding of how `socketio` currently works.

1. Who owns the port? .NET (Kestrel) or `main.js`?
2. How can we make sure that only our .NET process and the Electron process it launched can use the `socketio` channel?

FlorianRappl:
1. We need to see. Preferrably it's still Node.js - but with automatic port selection. The port would then either be set via arguments (Node -> .NET) or communicated via stdout (.NET -> Node).
2. The concept from this PR is still used.

epsitec:
I am still convinced that our approach with a single port allocated by Kestrel and filtered using a dedicated middleware is the only realistic solution.

FlorianRappl:
I think we are approaching the end here.

- SignalR is removed - the auth mechanism works with SocketIO
- Node.js *always* initiates the connection (incl. port selection); it also chooses the auth token
- The previous mechanism of expanding the auth token to the ASP.NET Core served page via some middleware is still present
- Authentication is always on (potentially we might want to opt-out of it, but I cannot think of good reasons for that - debugging in a normal browser should not be a good reason imho)
- There is no need for these free port finding codes / functions; it works by using `0` (i.e., OS functionality)

Right now I kept the Blazor sample but I can also remove it / put it in a dedicated PR. As this uses the new middleware I think keeping it in here (or otherwise making it a follow-up PR) is suitable.

docs/Dev/Authentication.md
```diff
@@ -0,0 +1,583 @@
+# Electron.NET SignalR Authentication Guide
```
softworkz, line 1: SignalR is removed?

src/ElectronNET.API/Runtime/Services/ElectronProcess/ElectronProcessActive.cs
```diff
@@ -157,18 +160,36 @@ protected override Task StopCore()
...
         private async Task StartInternal(string startCmd, string args, string directoriy)
         {
-            try
+            var tcs = new TaskCompletionSource();
+
+            void Read_SocketIO_Parameters(object sender, string line)
             {
-                await Task.Delay(10.ms()).ConfigureAwait(false);
+                // Look for "Electron Socket: listening on port %s at ..."
+                var match = extractor.Match(line);
```
softworkz, line 168: Parsing the console output with a RegEx? That's truly horrible...

src/ElectronNET.Host/main.js
```diff
@@ -73,7 +82,7 @@ let manifestJsonFilePath = path.join(currentPath, manifestJsonFileName);
 
 // if running unpackedelectron, lets change the path
 if (unpackedelectron || unpackeddotnet) {
-    console.log('unpackedelectron! dir: ' + currentPath);
+    console.debug('Running in unpacked mode, dir: ' + currentPath);
```
softworkz, line 85: Isn't this essentially removing that line from the output?

src/ElectronNET.Host/main.js
```diff
@@ -140,7 +149,7 @@ function getForwardedArgs() {
 
 const forwardedArgs = getForwardedArgs();
 
-app.on('ready', () => {
+app.on('ready', async () => {
```
softworkz: Why async? I see no awat.

src/ElectronNET.Host/main.js
```diff
@@ -153,44 +162,41 @@ app.on('ready', () => {
...
 });
 
 app.on('quit', async (event, exitCode) => {
-    try {
-        server.close();
-        server.closeAllConnections();
-    } catch (e) {
-        console.error(e);
+    // Clean up Socket.IO resources (legacy mode only)
+    if (typeof server !== 'undefined' && server) {
```
softworkz:
```suggestion
    if (server) {
```

src/ElectronNET.Host/main.js
```diff
@@ -153,44 +162,41 @@ app.on('ready', () => {
...
+            console.error('Error closing Socket.IO server:', e);
+        }
     }
 
-    try {
-        apiProcess?.kill();
-    } catch (e) {
-        console.error(e);
+    // Clean up API process (Socket.IO mode only)
+    if (typeof apiProcess !== 'undefined' && apiProcess) {
```
softworkz, line 185:
```suggestion
    if (apiProcess) {
```

src/ElectronNET.Host/main.js
```diff
@@ -153,44 +162,41 @@ app.on('ready', () => {
...
+            apiProcess.kill();
+        } catch (e) {
+            console.error('Error killing API process:', e);
+        }
     }
 
-    try {
-        if (io && typeof io.close === 'function') {
+    // Clean up Socket.IO connection
+    if (typeof io !== 'undefined' && io && typeof io.close === 'function') {
```
softworkz, line 194:
```suggestion
    if (io && io.close) {
```

src/ElectronNET.Host/main.js
```diff
@@ -153,44 +162,41 @@ app.on('ready', () => {
...
-    });
 });
 
 app.on('quit', async (event, exitCode) => {
-    try {
-        server.close();
-        server.closeAllConnections();
-    } catch (e) {
-        console.error(e);
+    // Clean up Socket.IO resources (legacy mode only)
```
softworkz, line 174: What is legacy here?

src/ElectronNET.Host/main.js
softworkz: Also see other occurrences below

src/ElectronNET.Samples.BlazorSignalR/Components/Layout/MainLayout.razor
softworkz, line 1: You said that SignalR is removed, but this sample is named ElectronSignalR..?

src/ElectronNET.Samples.BlazorSignalR/Components/_Imports.razor
```diff
@@ -0,0 +1,11 @@
...
+@using System.Net.Http.Json
+@using Microsoft.AspNetCore.Components.Forms
+@using Microsoft.AspNetCore.Components.Routing
+@using Microsoft.AspNetCore.Components.Web
+@using static Microsoft.AspNetCore.Components.Web.RenderMode
+@using Microsoft.AspNetCore.Components.Web.Virtualization
+@using Microsoft.JSInterop
+@using ElectronNET.Samples.BlazorSignalR
+@using ElectronNET.Samples.BlazorSignalR.Components
+@using ElectronNET.Samples.BlazorSignalR.Components.Layout
```
softworkz, line 11: Is the sample still using SignalR?

src/ElectronNET.sln
```diff
@@ -1,3 +1,4 @@
+﻿
```
softworkz: Blank line?

src/ElectronNET.sln
```diff
@@ -74,53 +75,151 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ElectronNET.Samples.Electro
...
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "ElectronNET.Samples.BlazorSignalR", "ElectronNET.Samples.BlazorSignalR\ElectronNET.Samples.BlazorSignalR.csproj", "{96870775-C2F7-48C4-88F0-074BE1A1D679}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
+		Debug|x64 = Debug|x64
+		Debug|x86 = Debug|x86
 		Release|Any CPU = Release|Any CPU
+		Release|x64 = Release|x64
+		Release|x86 = Release|x86
```
softworkz: Why do you want to add those solution configurations?

src/ElectronNET/build/package.template.json
softworkz: I thought, SignalR is removed?

src/ElectronNET/build/ElectronNET.LateImport.targets
```diff
@@ -397,6 +397,19 @@
...
+      <_PslSourceDist>$(ElectronSourceFilesPath)\node_modules\psl\dist</_PslSourceDist>
+      <_PslTargetDist>$(ElectronOutDir)node_modules\psl\dist</_PslTargetDist>
+    </PropertyGroup>
+    
+    <ItemGroup>
+      <_PslDistFiles Include="$(_PslSourceDist)\**\*.*" Condition="Exists('$(_PslSourceDist)')" />
+    </ItemGroup>
+    
+    <Message Importance="High" Text="Copying psl dist folder from source" Condition="@(_PslDistFiles->Count()) > 0" />
+    <Copy SourceFiles="@(_PslDistFiles)" DestinationFolder="$(_PslTargetDist)\%(RecursiveDir)" Condition="@(_PslDistFiles->Count()) > 0" />
```
softworkz: What is this supposed to do?

softworkz (changes requested):
I haven't looked at everything yet (e.g. not the auth stuff)

docs/Dev/Authentication.md
FlorianRappl: Kept the document for now (but removed the others) to see what to keep. I think it would need some major overhaul.

src/ElectronNET.API/Runtime/Services/ElectronProcess/ElectronProcessActive.cs
FlorianRappl: I don't think so. We control that output. If you have a better approach that is as readable bring it up.

src/ElectronNET.Host/main.js
FlorianRappl: It's unnecessary and easy to control.

src/ElectronNET.Host/main.js
FlorianRappl: Definitely a left over.

softworkz:
The commit history is not sane. Maybe it's better to start from a fresh develop branch and cleanly apply the changes one after another, like:

- Introduce ISocketConnection interface
- Change port handling
- Introduce authentication mechanism
- Add docs
- Add sample

That would also avoid stray changes (which still exist, but much better than before).

Regarding the port exchange, why does it need to use parsing command line output? 
When dotnet starts Electron, then it already has its port and informs the Electron side about that port via command line param. Can't the Electron side then just send a first websocket message to the dotnet port which tells dotnet about the Electron port?

src/ElectronNET.API/Runtime/Services/ElectronProcess/ElectronProcessActive.cs
FlorianRappl: My idea was to remove essentially everything except for the `IElectronAuthenticationService` and `ElectronAuthenticationMiddleware` part. But here I am eager to get feedback what seems relevant.

src/ElectronNET.Host/main.js
softworkz: How?

src/ElectronNET.API/Runtime/Services/ElectronProcess/ElectronProcessActive.cs
softworkz: See comment in main conversation

src/ElectronNET.Samples.BlazorSignalR/Components/_Imports.razor
FlorianRappl: It actually is not - it's just using the authentication.

src/ElectronNET.sln
FlorianRappl: I don't :)

softworkz: LOL

src/ElectronNET/build/package.template.json
FlorianRappl: It is - we might want to change this template in the future. For now let's keep it and manually sync it. Changed.

FlorianRappl:
> When dotnet starts Electron, then it already has its port and informs the Electron side about that port via command line param. Can't the Electron side then just send a first websocket message to the dotnet port which tells dotnet about the Electron port?

That is unfortunately not a good approach. We want to have a stable port selection - and the best approach is to just let the OS decide the port. To avoid race conditions we should pick `0`, but then the owner of the port (the part that opens it) also needs to do it. Therefore, since Node.js always initiates the IPC port we need a way to communicate the port back. When Node.js starts this is easy via the command args - the other way around requires some back talking.

> The commit history is not sane. Maybe it's better to start from a fresh develop branch.

I agree on this one. I'd actually separate this into multiple PRs:

- Introduce ISocketConnection interface
- Change port handling and added authentication mechanism
- Docs + sample

src/ElectronNET/build/ElectronNET.LateImport.targets
FlorianRappl: I honestly don't know. I'd leave that out - not sure what was the intention of the OP here.

softworkz:
> That is unfortunately not a good approach. We want to have a stable port selection - and the best approach is to just let the OS decide the port. 

Yes, that's what I mean:

- Electron sets up socket connection (with port 0)
- It retrieves the port that has been assigned, eg. 55555
- Dotnet starts Electron, with 55555 in the command line
- Electron is starting up and creates its own socket (with port 0)
- It retrieves the port that has been assigned, eg. 77777
- It sends a message to 55555 which indicates that its own port is 77777

src/ElectronNET.Host/main.js
FlorianRappl: 1. It does not remove the output (default)
2. If we ever don't want certain messages (i.e., debug messages) we can just control this easily (not only find it / but actually programmatically omit it)
3. The logging intent is better communicated (it's just for debug purposes)

softworkz: That makes sense of course.
I was just wondering whether log.debug() will still print to STDOUT? If it does, it's all good.

softworkz:
Overall, I think the auth, the ISocketConnection and better port handling are all good and valuable additions!
(just in case I would have created a different impression so far 😉)

src/ElectronNET/build/ElectronNET.LateImport.targets
softworkz: When someone needs such kinds of customization, they can simply add their own target which is then executed after "our" target, like this:

```xml
<Target Name="CopyPslDistFolder" AfterTargets="ElectronConfigureApp" DependsOnTargets="ElectronConfigureApp">

  <!-- Fix psl dist folder issue - copy from source if it exists -->
  <PropertyGroup>
    <_PslSourceDist>$(ElectronSourceFilesPath)\node_modules\psl\dist</_PslSourceDist>
    <_PslTargetDist>$(ElectronOutDir)node_modules\psl\dist</_PslTargetDist>
  </PropertyGroup>
  
  <ItemGroup>
    <_PslDistFiles Include="$(_PslSourceDist)\**\*.*" Condition="Exists('$(_PslSourceDist)')" />
  </ItemGroup>
  
  <Message Importance="High" Text="Copying psl dist folder from source" Condition="@(_PslDistFiles->Count()) > 0" />
  <Copy SourceFiles="@(_PslDistFiles)" DestinationFolder="$(_PslTargetDist)\%(RecursiveDir)" Condition="@(_PslDistFiles->Count()) > 0" />

</Target>
```

src/ElectronNET/build/package.template.json
softworkz: At a very early stage I was considering to make this package template overridable by allowing users to put their own version in the Properties folder of the project (similar to electron-builder.json).

But then I figured that many changes that people would do will break things, because all the logic is controlled by the MSBuild properties. Those properties are used in many different places and must match what's specified in package.json. 
When we would allow users to freely customize package.json, they will make changes which would break everything and then wonder and complain about it.

That's why I disregarded that idea eventually. 
And the better way to allow more customization is to add more MSBuild properties which allow making changes to those parts of package.json which are safe to change.

FlorianRappl: My point was not about end user customization but rather about the shared parts with the `Host` project (deps).

PR #1031, opened 2026-02-17
Add Documentation for Preload Script

### LINKED ISSUE #1028: Preload Script is Not Loaded
I'm trying to expose certain utility functions to the renderer process. Electron advises using a preload script and expose the relevant functions via the [contextBridge](https://www.electronjs.org/docs/latest/tutorial/tutorial-preload). Unfortunately, it seems the preload script is not called when specifying the file path via the `BrowserWindowOptions.WebPreferences.Preload` option during window creation. Furthermore, it seems the preload file is not checked at all because even proving an invalid file path won't lead to an exception (which I somewhat expected in that scenario).

* **NuGet Version**: 0.4.0
* **Electron Version**: 39.2.6
* **Electron Builder Version**: 26.0.12
* **Framework**: net10.0
* **Target**: win-x64 (specified via publish profile)
* **IDE**: Visual Studio 2026 18.1.1

Steps to Reproduce:
1. Create a new Blazor Web App, targeting `net10.0`
2. Add `ElectronNET.Core` and `ElectronNET.Core.AspNet` version 0.4.0 as NuGet packages
3. Add `builder.UseElectron(args, ElectronAppReady);` to the builder call and the following method for creating the window:
```cs
private static async Task ElectronAppReady()
{
    var options = new BrowserWindowOptions
    {
        Show = false,
        IsRunningBlazor = true,
        WebPreferences = new WebPreferences
        {
             Preload = Path.Combine(AppContext.BaseDirectory, "preload.js")
        }
    };
    var browserWindow = await Electron.WindowManager.CreateWindowAsync(options);
    browserWindow.OnReadyToShow += () => browserWindow.Show();
}
```
4. Create a file called `preload.js` in the root directory of the project with the following content (example from the [Electron docs](https://www.electronjs.org/docs/latest/api/web-utils#webutilsgetpathforfilefile)):
```js
const { contextBridge, webUtils } = require('electron');

contextBridge.exposeInMainWorld('electronApi', {
    getPathForFile: (file) => {
        return webUtils.getPathForFile(file);
    }
});
```
5. Add the following to the project configuration (`.csproj` file) to ensure the preload script is copied to the output directory:
```xml
<ItemGroup>
    <Content Include="preload.js" CopyToOutputDirectory="PreserveNewest" />
</ItemGroup>
```
6. Run the application and try to access `window.electronApi.getPathForFile` in JavaScript

AeonSake:
Proposed changes for #1028

Adds a check for a user-defined preload script before manually overriding the path when running in Blazor-mode. This is _technically_ a breaking change (since the behavior has changed) as preload scripts are now always executed. However, this most likely has no impact on existing applications as they would not have worked before this change anyway (i.e., when using a preload script in Blazor-mode).

### DIFF
--- a/src/ElectronNET.Host/api/browserWindows.ts
+++ b/src/ElectronNET.Host/api/browserWindows.ts
@@ -235,8 +235,8 @@ export = (socket: Socket, app: Electron.App) => {
       };
     }
 
-    if (options.isRunningBlazor) {
-      options.webPreferences["preload"] = path.join(
+    if (options.isRunningBlazor && !options.webPreferences.preload) {
+      options.webPreferences.preload = path.join(
         __dirname,
         "..",
         "scripts",
--- a/src/ElectronNET.Host/api/browserWindows.ts
+++ b/src/ElectronNET.Host/api/browserWindows.ts
@@ -235,8 +235,8 @@ export = (socket: Socket, app: Electron.App) => {
       };
     }
 
-    if (options.isRunningBlazor && !options.webPreferences.preload) {
-      options.webPreferences.preload = path.join(
+    if (options.isRunningBlazor) {
+      options.webPreferences["preload"] = path.join(
         __dirname,
         "..",
         "scripts",
--- a/docs/Using/Configuration.md
+++ b/docs/Using/Configuration.md
@@ -85,6 +85,26 @@ BrowserWindowOptions browserWindowOptions = new BrowserWindowOptions
 ```
 
 
+### Preload Script
+
+If you require the use of a [preload script](https://www.electronjs.org/docs/latest/tutorial/tutorial-preload), you can specify the file path of the script when creating a new window like so:
+
+```csharp
+WebPreferences wp = new WebPreferences();
+wp.Preload = "path/to/preload.js";
+BrowserWindowOptions browserWindowOptions = new BrowserWindowOptions
+{
+    WebPreferences = wp
+};
+```
+
+> [!IMPORTANT]
+> When using a preload script _AND_ running a Blazor app, the following lines must be added to the preload script:
+> ```js
+> global.process = undefined;
+> global.module = undefined;
+> ```
+
 
 ## 🚀 Next Steps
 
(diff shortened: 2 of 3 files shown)


src/ElectronNET.Host/api/browserWindows.ts
```diff
@@ -235,8 +235,8 @@ export = (socket: Socket, app: Electron.App) => {
       };
     }
 
-    if (options.isRunningBlazor) {
-      options.webPreferences["preload"] = path.join(
+    if (options.isRunningBlazor && !options.webPreferences.preload) {
```
FlorianRappl: Does not make sense as `isRunningBlazor` only exists for this.

The warning should be printed that a preload script will be ignored if `isRunningBlazor` is set. The option, which has been turned on explicitly, should always have preference.

FlorianRappl (changes requested):

src/ElectronNET.Host/api/browserWindows.ts
AeonSake: Fine by me, I'll update the note in the documentation to reflect this.

AeonSake:
This PR has been modified to only add new documentation to clarify how to handle preload scripts (and a note when using Blazor).

softworkz:
We could make `CreateWindowAsync()` throw an `ArgumentException` if `IsRunningBlazor` is true and `Preload` is specified.

AeonSake:
> We could make `CreateWindowAsync()` throw an `ArgumentException` if `IsRunningBlazor` is true and `Preload` is specified.

That would probably also be a good idea for user feedback, I can add that in another commit if wanted.

softworkz:
> > We could make `CreateWindowAsync()` throw an `ArgumentException` if `IsRunningBlazor` is true and `Preload` is specified.
> 
> That would probably also be a good idea for user feedback, I can add that in another commit if wanted.

That would be great.

I assume that by "user", you mean the "developer consuming Electron.NET". Of course this exception is meant to warn developers; it should never surface to end-users of the application. 
(it obviously won't, unless the developer would never test it 🤣 )

FlorianRappl (approved):
Let's merge it in right now - we can do that exception in a separate PR.

softworkz:
Yup - I didn't meant to block on that..

PR #1033, opened 2026-02-25
Add overload for GetAllDisplaysAsync() with timout parameter

softworkz:


softworkz:
I've merged this as it's trivial enough and I need a package.

PR #1037, opened 2026-03-05
Changed facade to socket connection

FlorianRappl:
Follow up for #1021: Introduces a new interface `ISocketConnection` and renames the facade to `SocketConnection`.

Exclusively uses the interfaces for the type.


softworkz (approved):
Very good!

PR #1038, opened 2026-03-05
Automatic Port Selection and Secured Communication

FlorianRappl:
Follow up for #1021 - should be reviewed / merged after #1037.

Introduces an authentication token to be used in the communication between Node.js (Electron) and .NET. The same token can be used to also guard an ASP.NET Core server - allowing only access from within the boundaries of the running application.

The port is now OS-selected - guaranteeing a free accessible port. No tools to find free ports are necessary.

The owning part (e.g., Node.js for the WebSocket connection to control the Electron application) also owns / opens the port - independent of the used startup mode.


FlorianRappl:
Should we move this forward @softworkz ?

softworkz:
Sorry, haven't come to this yet.

Why is there a merge conflict?

FlorianRappl:
> Sorry, haven't come to this yet.
> 
> Why is there a merge conflict?

There is none. There was initially one as planned due to the splitting.

softworkz:
image

image

FlorianRappl:
Shows different for me:

image

Also:

image

Furthermore:

```
$> git pull origin develop
From https://github.com/ElectronNET/Electron.NET
 * branch            develop    -> FETCH_HEAD
Already up to date.
$> git push origin feature/socket-enhancements
Everything up-to-date
```

src/ElectronNET.API/Runtime/Services/ElectronProcess/ElectronProcessActive.cs
```diff
@@ -157,18 +160,36 @@ protected override Task StopCore()
...
                 Console.Error.WriteLine("[StartInternal]: startCmd: {0}", startCmd);
                 Console.Error.WriteLine("[StartInternal]: args: {0}", args);
 
                 this.process = new ProcessRunner("ElectronRunner");
                 this.process.ProcessExited += this.Process_Exited;
+                this.process.LineReceived += Read_SocketIO_Parameters;
                 this.process.Run(startCmd, args, directoriy);
 
-                await Task.Delay(500.ms()).ConfigureAwait(false);
+                await tcs.Task.ConfigureAwait(false);
```
softworkz: If electron errors and doesn't print that line, the task will hang forever. 
Whether process will still exit, I'm not sure, but it should keep hanging their either case.

src/ElectronNET.Host/main.js
```diff
@@ -286,9 +290,16 @@ function startSocketApiBridge(port) {
...
     // @ts-ignore
     io.on('connection', (socket) => {
-        console.log('Electron Socket: connected!');
+        console.info('Electron Socket: connected!');
+
+        if (authToken && socket.request.headers.authorization !== authToken) {
+            console.warn('Electron Socket authentication failed!');
+            socket.disconnect(true);
+            return;
+        }
```
softworkz, line 324: This makes it look like the auth token would be optional, but it isn't
(yet it should be)

src/ElectronNET.API/Runtime/Services/ElectronProcess/ElectronProcessActive.cs
```diff
@@ -157,18 +160,36 @@ protected override Task StopCore()
...
+                {
+                    var port = int.Parse(match.Groups[1].Value);
+                    var token = match.Groups[2].Value;
+
+                    this.process.LineReceived -= Read_SocketIO_Parameters;
+                    ElectronNetRuntime.ElectronAuthToken = token;
+                    ElectronNetRuntime.ElectronSocketPort = port;
+                    tcs.SetResult();
+                }
+            }
```
softworkz: This class is always providing the `electronforcedport` parameter, in that case it's pointless to monitor the output for a port number.

src/ElectronNET.API/Runtime/Services/ElectronProcess/ElectronProcessActive.cs
```diff
@@ -157,18 +160,36 @@ protected override Task StopCore()
 
         private async Task StartInternal(string startCmd, string args, string directoriy)
         {
-            try
+            var tcs = new TaskCompletionSource();
+
+            void Read_SocketIO_Parameters(object sender, string line)
             {
-                await Task.Delay(10.ms()).ConfigureAwait(false);
+                // Look for "Electron Socket: listening on port %s at ..."
```
softworkz: There should be a path completely without reading the process output.
(without auth token)

src/ElectronNET.Host/main.js
```diff
@@ -73,7 +83,7 @@ let manifestJsonFilePath = path.join(currentPath, manifestJsonFileName);
 
 // if running unpackedelectron, lets change the path
 if (unpackedelectron || unpackeddotnet) {
-    console.log('unpackedelectron! dir: ' + currentPath);
+    console.debug('Running in unpacked mode, dir: ' + currentPath);
```
softworkz:
```suggestion
    console.debug('Running in unpackaged mode, dir: ' + currentPath);
```

softworkz (changes requested):
I'm very sorry for responding so late!

src/ElectronNET.API/Runtime/Services/ElectronProcess/ElectronProcessActive.cs
FlorianRappl: It should hang there as we can't communicate.

What we can talk about is a barrier here (e.g., 10 seconds) - after that the process would terminate. I would, however, not make it too small. Reason for not including that right now is that I am not sure how viable it is - and choosing a number that is too low might create more problems than solve here.

src/ElectronNET.Host/main.js
FlorianRappl: I would not make it optional - its part of the standard flow. There is no reason to have this optional.

src/ElectronNET.API/Runtime/Services/ElectronProcess/ElectronProcessActive.cs
FlorianRappl: I disagree. We should always monitor here. Right now the forcedport is like a recommendation - it can be also "0" / or later on use a different port in case the recommended one was blocked.

I would always follow the same logic here - makes it easier to debug and reason.

src/ElectronNET.API/Runtime/Services/ElectronProcess/ElectronProcessActive.cs
FlorianRappl: I would not do that. It's less secure and just more complicated. Let's keep it to the point.

src/ElectronNET.API/Runtime/Services/ElectronProcess/ElectronProcessActive.cs
softworkz: Just register the exit event to try-cancel the tcs in addition to a comfortable timeout

src/ElectronNET.API/Runtime/Services/ElectronProcess/ElectronProcessActive.cs
softworkz: I do not want to have that in our application that dotnet is relying on the console output of Electron.

softworkz: I have nothing against making the auth token mandatory, but it can be provided by the dotnet side as cli param, so it doesn't need to be read it from the console output. (that's no more and no less secure)

FlorianRappl: Theoretically possible, but that's another branch / logic that we would need to take care of. So I'd opt out of that.

src/ElectronNET.API/Runtime/Services/ElectronProcess/ElectronProcessActive.cs
FlorianRappl: Yes, I think this in combination with a 2-minute timeout should deal with all those cases.

I'll update the PR.

src/ElectronNET.API/Runtime/Services/ElectronProcess/ElectronProcessActive.cs
softworkz: Was just a thought. My only actual concern is the console parsing. That's admittedly a rather egoistic concern. I do not mean to say that parsing CLI output is generally wrong or bad, it's about the specific context of our app: Linux is a world of 1001 surprises, often bad surprises and I don't want to employ anything which has potential for causing bad surprises - that's why I would like to retain a path where this doesn't happen.

Thanks

FlorianRappl: Yeah can relate to that... My suggestion would be to have it merged / out in a preview version and test it extensively. If that does not hold / has weak points then I think a different approach must be anyway found (because I would still insist that a single flow is better than multiple - but that single flow has to work in all cases).

src/ElectronNET.API/Runtime/Services/ElectronProcess/ElectronProcessActive.cs
FlorianRappl: Updated.

src/ElectronNET.API/Runtime/Services/ElectronProcess/ElectronProcessActive.cs
softworkz: The problem is this: as soon as this is merged, there is no more point for me to submit any fixes and updates, because I cannot use the packages anymore.  

Beyond my own usage, introducing something new in a way that it's optional, is generally advantageous:

- Nobody's story will be hard broken or blocked
- It's useful for differential diagnosis when issues are reported, to determine whether it's by the new code or due to other reasons
- It would make me very happy 😉

GregorBiswanger:
I took over this PR and addressed the review concern around parsing Electron console output.

The dotnet-first startup flow now generates the auth token on the .NET side and passes it to Electron via environment variable. Electron reports the OS-selected socket port through a temporary startup-info file, so .NET no longer depends on parsing Electron's stdout. The auth token is no longer written to normal logs.

I also kept the Electron-first flow intact and removed logging of backend startup parameters because they include the auth token.

@softworkz I requested a re-review because your original concern should now be addressed by 9c89d84.

PR #1049, opened 2026-03-25
Fix `HasShadow` Not Being `true` by Default

### LINKED ISSUE #1048: `HasShadow` Is Not `true` by Default
The documentation of `BrowserWindowOptions.HasShadow` says it's `true` by default (mirroring what electron does). However, the class property does not force `true` by default and therefore gets serialized as `false` when creating a new window.

https://github.com/ElectronNET/Electron.NET/blob/c8268fae0eab5e6d1f79ae82c9b709ebb35b49aa/src/ElectronNET.API/API/Entities/BrowserWindowOptions.cs#L215-L218

AeonSake:
Fixes #1048

### DIFF
--- a/src/ElectronNET.API/API/Entities/BrowserWindowOptions.cs
+++ b/src/ElectronNET.API/API/Entities/BrowserWindowOptions.cs
@@ -215,7 +215,8 @@ namespace ElectronNET.API.Entities
         /// <summary>
         /// Whether window should have a shadow. Default is true.
         /// </summary>
-        public bool HasShadow { get; set; }
+        [DefaultValue(true)]
+        public bool HasShadow { get; set; } = true;
 
         /// <summary>
         /// Forces using dark theme for the window, only works on some GTK+3 desktop environments. Default is false.


FlorianRappl (approved):
LGTM.

Personally, I would rather change the default / documentation to `false`. This way, there is nothing needed and it works as right now. However, imho changing it is fine to align with Electron's default behavior (and the previous documentation - so no one should be surprised by this from a documentation POV).

AeonSake:
> Personally, I would rather change the default / documentation to `false`. This way, there is nothing needed and it works as right now. However, imho changing it is fine to align with Electron's default behavior (and the previous documentation - so no one should be surprised by this from a documentation POV).

Makes sense since the default behavior has now changed. Though I'd argue that the default behavior so far was not working as intended (i.e., not as documented or expected from electron defaults), so this is more like a bug fix than a change in my opinion.

AeonSake:
Also, I only noticed this because on macOS the windows don't have any shadows by default. On Windows it seems that the window frame always has a shadow, regardless of that property (at least on Windows 10). But that might also depend on the other window customization settings.

FlorianRappl (approved):

PR #1077, opened 2026-06-19
Add Overloads for `UseElectron`

### LINKED ISSUE #1076: Add Overloads for `UseElectron`
It would be great to have some method overloads for the `UseElectron` extension method, specifically the following:
```cs
UseElectron(string[] args, Func<IServiceProvider, Task> onAppReadyCallback)
UseElectron(string[] args, Func<string[], Task> onAppReadyCallback)
UseElectron(string[] args, Func<IServiceProvider, string[], Task> onAppReadyCallback)
```
Both, `WebApplicationBuilderExtensions` and `WebHostBuilderExtensions`, would need these overloads.

#### Motivation
The current implementation requires the callback task to be ready _before_ the app is instantiated, i.e., before the service collection is initialized. This can cause issues, especially when the callback comes from a service. For that, it would be great to have an overload that allows access to the `IServiceCollection`, similarly to how the service collection factory pattern works. This allows the normal dependency injection logic to fully resolve the required services on demand, rather than having it initialized beforehand. Furthermore, having access to the arguments at that stage rather than having to pipe them through yourself would also reduce friction.
From what I can tell, the internal `ElectronNetRuntime.OnAppReadyCallback` is not executed before the app is instantiated which means, the fully initialized service collection would be available at that time.

On a side-note: the argument array has to be provided on the initial `UseElectron` method, but is never actually used?

AeonSake:
Adds the new overloads proposed in #1076. Also replaces the previously static property with a resolver service.

### DIFF
--- a/src/ElectronNET.API/ElectronNetRuntime.cs
+++ b/src/ElectronNET.API/ElectronNetRuntime.cs
@@ -50,8 +50,6 @@
 
         internal static int? ElectronProcessId { get; set; }
 
-        internal static Func<Task> OnAppReadyCallback { get; set; }
-
         internal static ISocketConnection GetSocket()
         {
             return RuntimeControllerCore?.Socket;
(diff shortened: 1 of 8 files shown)


FlorianRappl (approved):
LGTM - thanks!

PR #1081, opened 2026-07-06
chore: use default sdk runtime identifier

epsnm:
* Evaluate current runtime identifier by using msbuild property function of interop services RuntimeInformation instead of fixed value.

Assisted-By: Claude:claude-5-sonnet


FlorianRappl (approved):
LGTM - thanks!

PR #1087, opened 2026-07-27
Bug/#1084 - Visual studio loading issue

epsnm:
* Fix visual studio loading issue
* Let electron-builder choose current platform when not in WSL cross-compile mode
* MSBuild configuration barriers to be included in projects with CPM


src/ElectronNET.Host/ElectronNET.Host.esproj
```diff
@@ -1,4 +1,6 @@
-<Project Sdk="Microsoft.VisualStudio.JavaScript.Sdk/1.0.3864779">
+<Project>
+  <!--<Sdk Name="Microsoft.VisualStudio.JavaScript.SDK" Version="1.0.3864779" />-->
+  <Sdk Name="Microsoft.VisualStudio.JavaScript.SDK" Version="1.0.5581896" />
```
FlorianRappl: Was this accidentally committed?

FlorianRappl (commented):
In general looks great, but please keep existing JS.Sdk and I am not sure I understand the reason for pulling so many files into the root directory. From my perspective these changes are unrelated to the runtime identifier issue, but maybe I am not seeing the full picture here.

src/ElectronNET.Host/ElectronNET.Host.esproj
epsnm: This was du to an issue with Visual Studio 2026 where the sdk version is not available.

Understood I'll remove this change.

epsnm:
After reading the tests results it seems only one macos test is in error (macos-15-intel API-30.4.0)

```
ElectronNET.IntegrationTests.Tests.BrowserWindowTests.Represented_filename_and_edited_flags()
[xUnit.net 00:00:35.49]         --- End of stack trace from previous location ---
  Failed Represented_filename_and_edited_flags [1 s]
  Error Message:
   System.TimeoutException : No response after 1000ms trying to retrieve value browserWindow.GetRepresentedFilenameAsync()
  Stack Trace:
     at ElectronNET.IntegrationTests.Tests.BrowserWindowTests.Represented_filename_and_edited_flags() in /Users/runner/work/Electron.NET/Electron.NET/src/ElectronNET.IntegrationTests/Tests/BrowserWindowTests.cs:line 248
```

I've didn't catch how it can be related to these changes. Did you have other tests-runs that sometime fail for no apparent reason ?

I'll investigate with a run-until-fail session on a similar macos version.

FlorianRappl:
No need for this. Macos tests / machines are unfortunately flaky. Quite often just re-running the test on that particular image will succeed. 

It's a different story if multiple/all are failing.

FlorianRappl (approved):
LGTM thanks!

PR #1092, opened 2026-07-28
Bug/#1091 - Fix CA1416 on AspNet project

epsnm:
- Avoid CA1416 error with the CLI during release build.
- Removed CA1416 platform compatibility warning from .editorconfig and all .csproj files, reverting to default handling.
- Eliminating redundant conditions and only suppressing specified warnings (excluding CA1416).

### DIFF
--- a/src/ElectronNET.AspNet/ElectronNET.AspNet.csproj
+++ b/src/ElectronNET.AspNet/ElectronNET.AspNet.csproj
@@ -13,23 +13,6 @@
     <SymbolPackageFormat>snupkg</SymbolPackageFormat>
     <Nullable>disable</Nullable>
     <RootNamespace>ElectronNET</RootNamespace>
-  </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|net6.0|AnyCPU'">
-    <NoWarn>1701;1702;4014;CS4014;CA1416;CS1591</NoWarn>
-  </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|net8.0|AnyCPU'">
-    <NoWarn>1701;1702;4014;CS4014;CA1416;CS1591</NoWarn>
-  </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|net10.0|AnyCPU'">
-    <NoWarn>1701;1702;4014;CS4014;CA1416;CS1591</NoWarn>
-  </PropertyGroup>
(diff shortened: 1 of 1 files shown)


epsnm:
@FlorianRappl We've been forced to keep CA1416 on API due to 285+ warnings on this project. 

To be fixed in another issue

src/ElectronNET.AspNet/API/ServiceCollectionExtensions.cs
```diff
@@ -27,7 +28,13 @@ public static IServiceCollection AddElectron(this IServiceCollection services)
                 .AddSingleton(_ => Clipboard.Instance)
                 .AddSingleton(_ => HostHook.Instance)
                 .AddSingleton(_ => PowerMonitor.Instance)
-                .AddSingleton(_ => NativeTheme.Instance)
-                .AddSingleton(_ => Dock.Instance);
+                .AddSingleton(_ => NativeTheme.Instance);
+
+#if IS_MACOS_BUILD
```
FlorianRappl: This does not work.

Why? What we have here is our lib. It is a managed lib. As such it can be build (and will be built) on any system. That is what ships to all platforms. But what we want here is different: We want that sb building an application for a certain platform only gets it there.

So you need to replace compile-time check / guard by runtime-check.

FlorianRappl (changes requested):
I don't think there need to be so many changes. The previous PR already had too many changes (shifting around files).

Don't just remove things like the editorconfig.

src/ElectronNET.AspNet/API/ServiceCollectionExtensions.cs
epsnm: Understood, unfortunately a runtime guard is not an option as my issue is at compile time. I can add some `#pragma` to disable, but I don't like this at all.

Thanks for your review; I do now understand that the managed lib is built once for all. 
Therefore, it was forced to be build *without* platform target (win/linux/mac) or CPU architecture (*always AnyCPU*).

As for testing, we're currently building directly the managed lib, our Electron host propagates the Platform variable up to the Electron's project which, combined with the `.editorconfig` enforcing CA1416 in error level, triggers the issue.

Is there any reason to only disable warnings for the AnyCPU architecture in the csproj? 

Otherwise, aligning it to `ElectronNET.API.csproj` (same NoWarn regardless of current target, configuration, or architecture) is the solution.

FlorianRappl: Good question. Usually these warnings should go away with an appropriate (runtime) guard (that's at least their job - the compiler respects the guards and narrows it).

I agree that a `#pragma` disable is unfortunate. A better solution would be appreciated.

> Is there any reason to only disable warnings for the AnyCPU architecture in the csproj?

Not that I'm aware of, but I think aligning it as you suggested may be the best solution.

FlorianRappl (approved):
LGTM - thanks!

PR #1098, opened 2026-08-10
fix(packaging): Restore cross-compile on same platform and packaging dir name behavior

epsnm:
* Architecture must be defined in electron-builder compile args in order to allow cross-compile on same platform. 
* Enforce default-arch to be the default arch in order to avoid arch in package directory name.


FlorianRappl (approved):
LGTM

PR #1106, opened 2026-08-24
Add DotNet-First packaging layout support and build extensibility

DYH1319:
## Summary

This PR makes the Electron.NET runtime and MSBuild targets support an additional "DotNet-First" packaged app layout and gives packaging builds a few more escape hatches.

### Motivation

Today the runtime probes for the packaged layout `resources/bin/<app>` (Electron-first). When the .NET publish output is the app root and the Electron host is placed under an `electron` subdirectory (DotNet-First), `UnpackagedDetector` mis-identifies the app as unpackaged and `ElectronProcessActive` cannot find the Electron binary.

In addition, `ElectronPackageId`, `Title`, the intermediate `PublishDir`, and the `npm`/`npx` exec steps are hard-coded, which makes it harder to plug Electron.NET into custom packaging pipelines or CI.

### Changes

- **DotNet-First layout detection** in `UnpackagedDetector.CheckUnpackaged2`:
  - If the base directory contains an `electron` subdirectory, treat the app as packaged.
  - Existing `resources/bin` (Electron-first) and `.electron` (development) checks are preserved.

- **DotNet-First binary startup** in `ElectronProcessActive.StartCore`:
  - When `dir/electron` exists, start `electron` from that directory and use it as the working directory.
  - Falls back to the existing `dir.Parent.Parent` logic for the Electron-first layout.

- **Overridable build properties** in `ElectronNET.Core.props`:
  - `ElectronPackageId` and `Title` only default from `MSBuildProjectName` when the properties are not already set.

- **Configurable publish/exec targets** in `ElectronNET.LateImport.targets`:
  - `ElectronPublishDir` can override the intermediate `PublishDir`.
  - `ElectronSkipExecCommands` (default `false`) can suppress `npm`/`npx` execution steps.
  - All `Exec` tasks for `npm`/`npx` now respect `ElectronSkipExecCommands`.

### Files changed

- `src/ElectronNET.API/Runtime/Helpers/UnpackagedDetector.cs`
- `src/ElectronNET.API/Runtime/Services/ElectronProcess/ElectronProcessActive.cs`
- `src/ElectronNET/build/ElectronNET.Core.props`
- `src/ElectronNET/build/ElectronNET.LateImport.targets`

### DIFF
--- a/src/ElectronNET.API/Common/ProcessRunner.cs
+++ b/src/ElectronNET.API/Common/ProcessRunner.cs
@@ -499,7 +499,7 @@
             try
             {
                 // This shouldn't throw here, but the mono process implementation doesn't always behave as it should.
-                this.process.WaitForExit();
+                this.process.WaitForExit(1000);
             }
             catch (Exception ex)
             {
--- a/src/ElectronNET.API/Runtime/Services/ElectronProcess/ElectronProcessActive.cs
+++ b/src/ElectronNET.API/Runtime/Services/ElectronProcess/ElectronProcessActive.cs
@@ -87,10 +87,24 @@
             }
             else
             {
-                dir = dir.Parent!.Parent!;
-                startCmd = Path.Combine(dir.FullName, this.electronBinaryName);
+                // DotNet-First packaging mode: the Electron host is in the electron subdirectory.
+                var electronDir = Path.Combine(dir.FullName, "electron");
+
+                if (Directory.Exists(electronDir))
+                {
+                    // New DotNet-First directory layout.
+                    startCmd = Path.Combine(electronDir, this.electronBinaryName);
+                    workingDir = electronDir;
+                }
+                else
+                {
+                    // Fallback to the legacy Electron-First directory layout.
+                    dir = dir.Parent!.Parent!;
+                    startCmd = Path.Combine(dir.FullName, this.electronBinaryName);
+                    workingDir = dir.FullName;
+                }
+
                 args = $"-dotnetpacked -electronforcedport={this.socketPort:D} " + this.extraArguments;
-                workingDir = dir.FullName;
             }
 
             // We don't await this in order to let the state transition to "Starting"
--- a/src/ElectronNET.API/Runtime/Helpers/UnpackagedDetector.cs
+++ b/src/ElectronNET.API/Runtime/Helpers/UnpackagedDetector.cs
@@ -64,11 +64,20 @@
         private static bool? CheckUnpackaged2()
         {
             var dir = new DirectoryInfo(AppDomain.CurrentDomain.BaseDirectory);
+
+            // Electron-First packaging mode: the .NET binary is in resources/bin.
             if (dir.Name == "bin" && dir.Parent?.Name == "resources")
             {
                 return false;
             }
 
+            // DotNet-First packaging mode: the Electron host is in the electron subdirectory.
+            if (dir.GetDirectories().Any(e => e.Name == "electron"))
+            {
+                return false;
+            }
+
+            // Development mode: the .electron subdirectory exists.
             if (dir.GetDirectories().Any(e => e.Name == ".electron"))
             {
                 return true;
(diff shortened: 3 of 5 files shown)


src/ElectronNET.API/Common/ProcessRunner.cs
```diff
@@ -499,7 +499,7 @@ private void WaitForExitAfterExited()
             try
             {
                 // This shouldn't throw here, but the mono process implementation doesn't always behave as it should.
-                this.process.WaitForExit();
+                this.process.WaitForExit(1000);
```
FlorianRappl, line 502: I really don't like having a number here - if we do it should be configurable (and the default should be at relatively high, at least 5_000, i.e., smaller than the indefinite one we have right now, but still high).

src/ElectronNET.API/Runtime/Services/ElectronProcess/ElectronProcessActive.cs
```diff
@@ -87,10 +87,24 @@ protected override async Task StartCore()
...
+
+                if (Directory.Exists(electronDir))
+                {
+                    // New DotNet-First directory layout.
+                    startCmd = Path.Combine(electronDir, this.electronBinaryName);
+                    workingDir = electronDir;
+                }
+                else
+                {
+                    // Fallback to the legacy Electron-First directory layout.
```
FlorianRappl, line 101: This should not be "legacy". It should be the primary. Why do we have two different paths here? Can't this be just configurable ("electron-root-dir" - default is ".")?

FlorianRappl (commented):
I miss a bit why we need that extra complexity and what problem this solves.

In addition, to accept this PR we also need changes to the docs (how to activate, when to use, ...).

DYH1319:
Thanks for the review.

> I miss a bit why we need that extra complexity and what problem this solves.

Electron.NET's documentation already supports packaged .NET-first startup: the .NET executable starts first and then launches Electron. However, in the current code, when packaged startup runs, ElectronProcessActive assumes the .NET executable is under resources/bin, so it uses dir.Parent.Parent to walk up two directories to locate Electron.

Our packaging scenario looks like this:


```text
<install-root>/
  MyApp.exe              # .NET executable
  electron/
    MyApp.exe            # Electron executable
    resources/
    locales/
    ...
```

In this layout, the .NET executable sits at the package root, and Electron is placed in an electron/ subdirectory. At this point, dir.Parent.Parent jumps outside the package root, so it cannot find Electron. UnpackagedDetector also misidentifies the app as unpackaged because it doesn't see resources/bin.

So the core goal of this PR is: let the .NET runtime know where to find Electron, so that a .NET-first layout with the .NET entry point at the package root can actually work.

> In addition, to accept this PR we also need changes to the docs (how to activate, when to use, ...).

I'll change the docs to reflect the changes.

src/ElectronNET.API/Common/ProcessRunner.cs
DYH1319: Agreed. 1000ms was arbitrary. I'll make it configurable with a safer default of at least 5000ms.

src/ElectronNET.API/Runtime/Services/ElectronProcess/ElectronProcessActive.cs
DYH1319: Agreed. The existing resources/bin layout is still the primary layout in Electron.NET, and it shouldn't be called "legacy". I'll change this to an explicit configuration, e.g. an ElectronRootDir property. The default behavior will preserve the existing dir.Parent.Parent logic; when explicitly set, the path will be resolved relative to the .NET executable directory.

FlorianRappl:
Understood - my point was that the default packaging is Electron-first - so while you can always "compose arbitrarily to use the different modes" I was wondering if this affects / would be good to extend deployment / package building as documented at https://github.com/ElectronNET/Electron.NET/wiki/Package-Building

Let's refine the two things / make them configurable and then merge this. Many thanks for your efforts!

FlorianRappl:
Cherry picked to `develop`.

PR #1107, opened 2026-08-27
Fix/#1105 - Upgrade SocketIO dependency and disable proxies.

### LINKED ISSUE #1105: Electron app does not start correctly in case of proxies
* **Version**: API 0.5.2
* **.NET**: 10.0.302 SDK
* **Target**: win-x64 (Windows 11)

The socket.io bridge between the Electron host and the .NET backend runs over
`http://localhost:<electronPort>`, but the .NET client never opts out of the system
proxy. On Windows, as soon as a proxy is configured without `<local>` in the bypass
list, the engine.io handshake is sent to the proxy instead of the loopback socket
server, `ConnectAsync` never completes, and the app never opens a window.

Steps to Reproduce:

1. Windows → Settings → Network & Internet → Proxy → *Manual proxy setup*: enable it,
   set any address (e.g. `192.168.1.56:80`), and leave the exception list **empty**
   (i.e. do not tick "Don't use the proxy server for local addresses").
2. Start any Electron.NET app in Electron-first mode.
3. Electron logs `Electron Socket: listening on port <p> at 127.0.0.1`, the backend
   process starts, but `BridgeConnector connected!` is never printed and no window
   is created. The app hangs on the splash screen (or exits silently if none).
4. Re-tick "Don't use the proxy server for local addresses" (or disable the proxy)
   and restart → the app starts normally.

**Cause**

`SocketBridgeService` builds the loopback URL:

```csharp
// src/ElectronNET.API/Runtime/Services/SocketBridge/SocketBridgeService.cs:22
this.socketUrl = $"http://localhost:{this.socketPort}";
```

and `SocketIOConnection` passes options that never set `Proxy`:

```csharp
// src/ElectronNET.API/Bridge/SocketIOConnection.cs:19-31
var opts = string.IsNullOrEmpty(authorization) ? new SocketIOOptions() : new SocketIOOptions
{
    ExtraHeaders = new Dictionary<string, string> { ["authorization"] = authorization },
};
_socket = new SocketIO(uri, opts);
```

SocketIOClient 3.1.2 defaults are `Transport = Polling`, `AutoUpgrade = true`,
`Proxy = null`. Both `IHttpClient.SetProxy` and `IClientWebSocket.SetProxy` are only
invoked when `Options.Proxy` is non-null, so:

* the polling handshake goes through `HttpClientHandler` with `UseProxy = true,
  Proxy = null` → falls back to `HttpClient.DefaultProxy` (WinINET settings);
* the websocket upgrade goes through `ClientWebSocket` with the same default.

.NET exempts loopback only when the Windows override list contains `<local>`:

```
ProxyOverride empty  (BypassOnLocal=false) -> http://192.168.1.56/
with <local>         (BypassOnLocal=true)  -> http://127.0.0.1:53123/
```

Chromium is unaffected because it has an implicit proxy bypass for loopback, so the
`BrowserWindow` side would work — the failure is entirely on the .NET client leg.

**Suggested fix**

Force a direct connection for the bridge only, so that application traffic keeps
honouring the user's proxy:

```csharp
var opts = new SocketIOOptions
{
    // The bridge is a loopback IPC channel, never real network traffic.
    Proxy = DirectProxy.Instance,
};

private sealed class DirectProxy : IWebProxy
{
    public static readonly DirectProxy Instance = new DirectProxy();
    public ICredentials Credentials { get; set; }
    public Uri GetProxy(Uri destination) => destination;
    public bool IsBypassed(Uri host) => true;
}
```

`IsBypassed => true` covers both the polling transport and the websocket upgrade.
Setting `HttpClient.DefaultProxy` globally would not be acceptable, since it would
also disable the proxy for the host application's own outbound requests.

```

Chromium is unaffected because it has an implicit proxy bypass for loopback, so the
`BrowserWindow` side would work — the failure is entirely on the .NET client leg.

**Suggested fix**

Force a direct connection for the bridge only, so that application traffic keeps
honouring the user's proxy:

```csharp
var opts = new SocketIOOptions
{
    // The bridge is a loopback IPC channel, never real network traffic.
    Proxy = DirectProxy.Instance,
};

private sealed class DirectProxy : IWebProxy
{
    public static readonly DirectProxy Instance = new DirectProxy();
    public ICredentials Credentials { get; set; }
    public Uri GetProxy(Uri destination) => destination;
    public bool IsBypassed(Uri host) => true;
}
```

`IsBypassed => true` covers both the polling transport and the websocket upgrade.
Setting `HttpClient.DefaultProxy` globally would not be acceptable, since it would
also disable the proxy for the host application's own outbound requests.

epsnm:
* Upgrade SocketIO dependency
* Disable proxies

Co-Authored-By: Claude Sonnet 5


src/ElectronNET.API/ElectronNET.API.csproj
```diff
@@ -27,9 +27,10 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
     </PackageReference>
-    <PackageReference Include="SocketIOClient" Version="3.1.2" />
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="10.0.3" />
+    <PackageReference Include="SocketIOClient" Version="4.0.5" />
     <PackageReference Include="System.Drawing.Common" Version="8.0.22" />
-    <PackageReference Include="System.Text.Json" Version="8.0.6" Condition=" '$(TargetFramework)' == 'net6.0' " />
+    <PackageReference Include="System.Text.Json" Version="10.0.3" Condition=" '$(TargetFramework)' == 'net6.0' " />
```
FlorianRappl, line 33: Why this increase/change? Should not be.

src/ElectronNET.API/ElectronNET.API.csproj
```diff
@@ -27,9 +27,10 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
     </PackageReference>
-    <PackageReference Include="SocketIOClient" Version="3.1.2" />
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="10.0.3" />
```
FlorianRappl, line 30: Should not be necessary - like the changes using it.

src/ElectronNET.API/Bridge/SocketIOConnection.cs
```diff
@@ -4,33 +4,67 @@ namespace ElectronNET.API;
...
 
 internal class SocketIOConnection : ISocketConnection
 {
     private readonly SocketIO _socket;
     private readonly object _lockObj = new object();
+    private readonly CancellationTokenSource _connectCts = new CancellationTokenSource();
     private bool _isDisposed;
 
-    public SocketIOConnection(string uri, string authorization)
+    public SocketIOConnection(string uri, string authorization, Action<IServiceCollection> configureSocketIO = null)
```
FlorianRappl, line 25: Everything that references IServiceCollection should be reverted.

There is no need (and architectural no gain) forwarding the service collection or an action using it.

If we take options (e.g., when configured in a DI) we should be able to resolve this much earlier (in a DI setup) and then forward accordingly.

FlorianRappl (commented):
In general this is fine, but:

- ElectronNET.API should have nothing to do with DI
- It should not be necessary to make as many changes we've seen here

I think taking this a bit more lightweight would benefit the PR.

FlorianRappl:
I'll close this for now. The SocketIO dependency-update should not require so many changes and the proxy is already disabled in `develop`.

